PR #3, opened 2013-06-11
Upgraded hydra and sufia

jcoyne:


mjgiarlo:
I think it should, @jeremyf.

:+1:

PR #4, opened 2013-07-29
[move_bewteen_collections] Changes to enable 'move to collection' feature

flyingzumwalt:
You might not actually want to merge this into scholarsphere, since the feature is actually aimed more for archivesphere, but either way, this shows what changes you need to make in order to enable the feature in scholarsphere/archivesphere.  Note: this relies on changes in hydra-collections reflected here: https://github.com/projecthydra/hydra-collections/pull/13/commits

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -18,10 +18,14 @@ class CollectionsController < ApplicationController
   include BlacklightAdvancedSearch::ParseBasicQ
   include BlacklightAdvancedSearch::Controller
   include Sufia::Noid # for normalize_identifier method
+  include Hydra::Collections::SelectsCollections
+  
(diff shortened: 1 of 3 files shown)


carolyncole:
@mjgiarlo, @DanCoughlin Is this something we want in SS?

mjgiarlo:
@cam156 On one hand, I could see this being used, but there's a difference between ASphere and SSphere when it comes to collection membership in that the former wants single ownership and the latter is many-to-many. So the notion of moving a file from one collection into another may not be the right metaphor for SSphere.  This is a useful bit of code to archive, maybe put in a redmine ticket, if we decide to do something similar for SSphere, though?

mjgiarlo:
Thoughts, @DanCoughlin ?

mjgiarlo:
Pulled in here: https://scm.dlt.psu.edu/issues/8963

Closing.

Issue #11, opened 2014-02-26
Test Automation Static Pages


carolyncole:
### All static pages work whether logged in or not
1. Select the About page by clicking the About link on the top right of the page
2. Test all hyperlinks throughout the text on the page to confirm they are working
3. Select the Help page by clicking the Help link on the top right of the page
4. Test the video, hyperlinks, and drop menus throughout the text on the page to confirm they are working (Remember the hyperlinks within the drop menus)
5. Select the Contact page by clicking the Contact link on the top right of the page
6. Test all hyperlinks on the page to confirm they are working
7. Test the Contact Form, fill in all fields:
   1. Select an Issue Type
   2. Enter your name
   3. Enter your email
   4. Enter a subject
   5. Enter a message
   6. Click Send
      1. Once sent you will (as tester) receive three emails:
         1. Thank you for contacting us 
         2. ‘ScholarSphere Redmine’ copy of form submitted
         3. ‘ScholarSphere Form’ copy of form submitted

PR #16, opened 2014-03-07
Test automation static pages

cswalina:
@cam156 

Here is a feature test ( refs issue #11 ) for "Test Automation Static Pages".

Assuming it meets your approval, how do you want us to add it to the develop branch? We were thinking that the cleanest way to do it would be to save the spec file, eliminate this branch, and then add the spec file to develop with a single commit (rather than merging in all of our scratch work commits in this branch). Does that sound like a good approach?

### DIFF
--- a/spec/features/generic_file_show_spec.rb
+++ b/spec/features/generic_file_show_spec.rb
@@ -23,7 +23,8 @@ describe "Showing the Generic File" do
       @gf2.save!
   end
   after(:all) do
-    @gf1.destroy  rescue puts "error occured destroying object"
+    @gf1.destroy
+    @gf2.destroy
   end
 
   before do
@@ -123,7 +124,7 @@ describe "Showing the Generic File" do
   def check_page(link_name)
       page.should have_link link_name
       click_on link_name
-      page.should have_content @gf1.title.first 
-      page.should_not have_content @gf2.title.first 
+      page.should have_content @gf1.title.first
+      page.should_not have_content @gf2.title.first
   end
 end
--- a/spec/features/static_pages_new_spec.rb
+++ b/spec/features/static_pages_new_spec.rb
@@ -1,0 +1,121 @@
+require 'spec_helper'
+
+describe "Static pages" do
+
+  let(:user) { FactoryGirl.find_or_create(:user) }
+
+  shared_examples "verifying each page" do |logged_in|
+    before { sign_in user } if logged_in
+
+    [
+        '/about',
+        '/help',
+        '/contact'
+    ].each do |path|
+      describe "The '#{path}' page" do
+        it "has verified hyperlinks" do
+          verify_links(path)
+        end
+        unless logged_in
+          # Verifying youtube links is done via visiting an external url.
+          # We can only do so if we are not logged in as a user.
+          it "has verified videos exist" do
+            verfiy_youtube_links(path)
+          end
+        end
+      end
+    end
+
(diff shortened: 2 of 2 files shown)


mjgiarlo:
An alternative approach that we use quite often in the Hydra project would be for you to squash your commits on this branch then force-push back to the same branch. (That way, up above it will say there's 1 commit rather than 11 commits.) Notes on that here: http://ndlib.github.io/practices/one-commit-per-pull-request/

Once you do that, this pull request will be automatically updated.  Once that's done and Travis comes back with good news, any of us can merge it.  Thanks!

cswalina:
Thanks Mike. We squashed and force pushed, but one of our new specs fails on Travis (all tests pass locally though). We'll have to dig into what's going on on Travis ...

mjgiarlo:
Weird.  I have no idea why adding a spec would cause a failure in the proxy spec.  Good luck tracking that down!

woods:
@cam156 

We need some help: this test is complete and passes individually, but the build fails intermittently on Travis. We're not having any luck tracking down what could be causing the failures elsewhere in the suite. 

We suspect that it's because of an incomplete setup/teardown somewhere that's causing the test ordering to become significant, but we don't have enough background knowledge to track this down efficiently.

Whenever it's convenient, can someone please do the following for us:
1. Take a quick look at our changes to make sure we're not missing a set up or tear down step that could be causing state problems.
2. Check out the branch and verify that the test passes for you.
3. See if you can provide any insight on why other tests intermittently fail on Travis when we introduced this test.

In the meantime, we're working on other branches, so this isn't exactly holding us up, but we can't complete it without some additional insight, if nothing else to confirm that we're not missing anything and it would be time well spent for us to investigate further.

Thanks!

mjgiarlo:
I'll take a quick look, @woods.

woods:
Thanks!

On Tue, Mar 11, 2014 at 5:42 PM, Michael J. Giarlo <notifications@github.com

> wrote:
> 
> I'll take a quick look, @woods https://github.com/woods.
> 
> ## 
> 
> Reply to this email directly or view it on GitHubhttps://github.com/psu-stewardship/scholarsphere/pull/16#issuecomment-37352752
> .

## 

Scott Woods
West Arete
http://westarete.com/
scott@westarete.com
814-753-4951

mjgiarlo:
And then GitHub went down.  Now that it's back, I'll take a quick look. :smile:

mjgiarlo:
I can't get that particular test to fail, but I do get a failure unrelated to your changes (and possibly related to setup/teardown). Going to troubleshoot this a bit, @woods.

mjgiarlo:
I pushed a commit to this branch with one minor change.  I'm not confident that it fixes anything but my tests pass now, FWIW.

mjgiarlo:
Well, that run passed anyway.  I think this is good to merge.  Thanks, y'all!

:clap:

woods:
Thank you so much, Mike!

On Tue, Mar 11, 2014 at 8:03 PM, Michael J. Giarlo <notifications@github.com

> wrote:
> 
> Merged #16 https://github.com/psu-stewardship/scholarsphere/pull/16.
> 
> ## 
> 
> Reply to this email directly or view it on GitHubhttps://github.com/psu-stewardship/scholarsphere/pull/16
> .

## 

Scott Woods
West Arete
http://westarete.com/
scott@westarete.com
814-753-4951

PR #20, opened 2014-05-06
added object type facet #9198

bmaddy:
If someone could close the store when this merges, I'd appreciate it (I don't have access).

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -33,7 +33,7 @@ ActiveRecord::Schema.define(version: 20140428172016) do
     t.datetime "updated_at"
   end
 
-  add_index "checksum_audit_logs", ["pid", "dsid"], name: "by_pid_and_dsid", using: :btree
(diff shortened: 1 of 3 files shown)


db/schema.rb
```diff
@@ -33,7 +33,7 @@
     t.datetime "updated_at"
   end
 
-  add_index "checksum_audit_logs", ["pid", "dsid"], name: "by_pid_and_dsid", using: :btree
+  add_index "checksum_audit_logs", ["pid", "dsid"], name: "by_pid_and_dsid"
```
carolyncole, line 36: Are these changes needed for the PR?

carolyncole:
does close the store mean close the ticket?

db/schema.rb
bmaddy, line 36: Nope. I thought the general practice was to check in any `schema.rb` changes though. Is that not the case here?

bmaddy:
Yep, just a typo. I meant "story".

PR #22, opened 2014-05-29
New dashboard

awead:
New dashboard features in Sufia.


app/views/dashboard/index.html.erb
```diff
@@ -0,0 +1,53 @@
+<%= content_for :heading do %>
```
mjgiarlo: What do we get from overriding this file in ScholarSphere?

spec/controllers/my/files_controller_spec.rb
```diff
@@ -0,0 +1,173 @@
+# Copyright © 2012 The Pennsylvania State University
```
mjgiarlo: You can remove these copyright statements.  (We used to do this and then Hydra changed its IP practices.)

spec/features/dashboard_files_spec.rb
```diff
@@ -0,0 +1,3 @@
+require 'spec_helper'
+
+
```
mjgiarlo: Do we need code in this file?  :D

spec/controllers/my/files_controller_spec.rb
awead: For just this file?

mjgiarlo: Pretty much across the board!

awead: There are 125 files with that statement in them.  To be clear, you want that one line removed from all those files?

app/views/dashboard/index.html.erb
awead: This is the display for the dashboard.  If I don't include the heading, it doesn't show up.  It probably would have been better to put this in a partial, but I was just replicating the content_for usage in sufia:

https://github.com/projecthydra/sufia/blob/master/app/views/layouts/sufia-dashboard.html.erb

spec/controllers/my/files_controller_spec.rb
mjgiarlo: For this PR, I'd say remove this entire copyright statement/Apache license from this file (lines 1-14) only.  We can nuke the other ones opportunistically as we go.

app/views/dashboard/index.html.erb
mjgiarlo: I guess I was just wondering why this isn't already in Sufia, applying the principle of "only override Sufia when it's absolutely necessary."  What are we doing here that's different from Sufia's dashboard?  (My comment was on this entire file, not just the first line, sorry!)

spec/controllers/my/files_controller_spec.rb
awead: Can we file an issue or a ticket in Redmine to remind us?

mjgiarlo: I'll hotplate it now.  Thx!

app/views/dashboard/index.html.erb
awead: Oh, I see.  Well, at the moment, we're not changing anything, but we will be shortly when I add the additional bits such as proxy and transfers that aren't in Sufia.  I can remove the file for this PR and then add the overridden one back in later, if that helps.

spec/controllers/my/files_controller_spec.rb
mjgiarlo: It's in Redmine as #9316.

app/views/dashboard/index.html.erb
mjgiarlo: It might be worth pinging Jim Coble at Duke to see when he plans to finish work on porting proxies/transfers over to Sufia.  If that is soon, I think you can remove this file from this PR.  If it's not gonna be a while, we should probably leave it here _and_ put in a Redmine ticket to the tune of, "remove dashboard override and proxy/transfer functionality from ScholarSphere when feature is ported to Sufia"?

awead: I would prefer removing the file and merging the PR.

spec/controllers/my/files_controller_spec.rb
```diff
@@ -0,0 +1,159 @@
+require 'spec_helper'
```
carolyncole, line 1: Is this test needed in scholarsphere?  Are we overriding things in the controller?

awead, line 1: These tests were originally in a controller since replaced by My::FilesController.  Retaining them here is to ensure integration.  They could be moved to features, but that should be another ticket, or part of a large process of reviewing test coverage.

mjgiarlo, line 1: I recommend leaving it in for now, and we can deal with the Sufia/Sphere duplication in Redmine#3733.

carolyncole, line 1: Ahh! Makes sense.  I wonder if we should be retaining them though.  There should be feature tests written for the new dashboard.  The existing "dashboard"/file listing "should" already have coverage.  Maybe remove the tests and generate a ticket for verifying the File List feature tests?

mjgiarlo:
:shipit:

PR #23, opened 2014-06-25
Updating to latest Sufia with fixes, refs #9378 #9312

awead:
@mtribone This commit removes a ton of css files that weren't being used... at least as far as I can tell.  The files were there but either their `require` statements were commented-out or just weren't there at all. Since there was no `require_tree` statement, they weren't getting compiled into the asset pipeline.

Other than than, this brings us up-to-date with the latest Sufia and fixes some responsiveness in the header.


mtribone:
@awead  Thanks, dude. Yeah there were a bunch pointed out by @flyingzumwalt and were deleted locally but never updated. Some of those may be for the marketing landing page, so we'll need to double check.

awead:
@mjgiarlo Want to merge? I think the best thing is to get this out to QA and get some feedback.

mjgiarlo:
:+1:  Thanks!

PR #26, opened 2014-07-29
removed font and table options from tinymce

bmaddy:


bmaddy:
I didn't run the tests locally for this one (since I need to leave before they'll finish). We should make sure travis is happy before merging.

mjgiarlo:
:+1:

PR #32, opened 2014-08-12
Tweaks to button layout and reducing to one color theme, refs #9421

mtribone:
Created SCSS partials for ScholarSphere theme and tweaked a bit of HTML for smaller screen sizes.


app/views/dashboard/_index_partials/_heading_actions.html.erb
```diff
@@ -0,0 +1,26 @@
+<div class="col-xs-12 heading-row">
```
mjgiarlo, line 1: What are we overriding in this file?  Should these changes be in Sufia?

mjgiarlo:
We're overriding a good number of Sufia's views here.  Should these changes, or a subset of them, be done in Sufia instead so that our upgrade path doesn't become more brittle?  I get overriding Sufia's views for features that only exist in ScholarSphere, like proxies/transfers, but I'm not sure why we need these view overrides.

mtribone:
@mjgiarlo When I first discussed this work with some members of the team, they felt that doing the work in ScholarSphere would help us move forward more quickly with our deadline–code freeze. The overrides are to refactor the previous "wireframe" model for one that "pops" and it required a little tweaking for using Bootstrap panels and such and some grid tweaks for display issues.

app/views/dashboard/_index_partials/_heading_actions.html.erb
mtribone, line 1: Hmmm. You know...I don't think that anything is being overridden. Gonna look again.

mtribone, line 1: Ah! I needed to make a block so that you could hover over the icon AND text for one link. Previously, only the text was an actionable item.

mjgiarlo, line 1: Ah, OK.

mjgiarlo:
For sure it'll help us move forward more quickly.  I'm OK with intentionally creating technical debt as long as we have a plan for paying it down (and giving back to the community, as I'm guessing they'll want Sufia to have a UI that pops) -- maybe that's as simple as "let's ticket items like this in Redmine and devote a post-2.0 launch sprint to moving sphere stuff back into Sufia."

I wonder what @jcoyne thinks about how mergeable the above would be if moved over to Sufia.

mtribone:
@mjgiarlo Yeah, man. I have a tan, vermilion, teal, yellow dashboard  theme ready to go that would make Sufia look awesome. Yes to planning and yes to seeing how merge-able these changes are–sans the color scheme.

mjgiarlo:
Why sans the color scheme?

OK, so will you put in the Redmine ticket to shift this over to Sufia?  Or do you want to hear from @jcoyne first?

I'm not trying to drag this out, really -- I just don't want us to make headaches for ourselves later unless we really need to.

app/views/dashboard/_index_partials/_heading_actions.html.erb
carolyncole, line 1: @mjgiarlo  We also have transfers that do not exist in Sufia

carolyncole, line 1: nm, that is a different partial..

mtribone:
@mjgiarlo Sans the Penn State color scheme, but with another color scheme.

carolyncole:
@mjgiarlo I'm definitely with @mtribone on this.  He is busy and we need to hit the deadline.  If you would like to refactor this PR into Sufia this week before we pin feel free, but the rest of us are working accessibility, which takes a higher priority.  How about we add a ticket into Sufia for the updates?  Do we have a succinct list of what this fixed to describe the ticket? Or we could just point at the commit...

mjgiarlo:
I think it'd be A-OK to add a ticket into Sufia, as long as we prioritize coming back to it at some point before it gets stale.  I think pointing at the PR URL is perfectly fine.

PR #38, opened 2014-10-29
Capistrano3

pcrum:
Branch adds the ability for final release to maintain itself as a git repository versus utilizing git archive to deploy the final release.

Also added 'rbenv_install' capistrano plugin which includes the ability to install rbenv, bundle, and ruby.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -37,6 +37,7 @@ set :whenever_identifier, -> {"#{fetch(:application)}_#{fetch(:stage)}"}
 
 # git for source control
 set :scm, :git
+set :git_strategy, Capistrano::Git::SubmoduleStrategy
 
 # Default value for :format is :pretty
 set :format, :pretty
@@ -142,14 +143,5 @@ namespace :deploy do
   end
  end
  after :symlink_shared, :remove_resque
-
- # Restart the application
- desc 'Restart application'
- task :restart do
-  on roles(:app), in: :sequence, wait: 5 do
-      execute :touch, release_path.join('tmp/restart.txt')
-  end
- end
- #after :publishing, :restart 
  after :restart, "passenger:warmup" 
 end
--- a/lib/capistrano/submodule_strategy.rb
+++ b/lib/capistrano/submodule_strategy.rb
@@ -1,0 +1,21 @@
+class Capistrano::Git < Capistrano::SCM
+  module SubmoduleStrategy
+    include DefaultStrategy
+
+    def test
+      test! " [ -d #{repo_path}/.git ] "
+    end
+ 
+    def clone
+      git :clone, '-b', fetch(:branch), '--recursive', repo_url, repo_path
+    end
+ 
+    def release
+      context.execute :rm, '-rf', release_path
+      git :clone, '--branch', fetch(:branch),
+        '--recursive',
+        '--no-hardlinks',
+        repo_path, release_path
+    end
+  end
+end
--- a/Capfile
+++ b/Capfile
@@ -20,6 +20,9 @@ require 'capistrano/rails' #rails (includes bundler, rails/assets and rails/migr
 require 'whenever/capistrano' #whenever 
 require 'capistrano/passenger' #passenger
 require 'capistrano-resque'  #resque
+require 'capistrano/git' #git
+require './lib/capistrano/submodule_strategy' #custom submodule strategy
+require 'capistrano/rbenv_install' #rbenv install plugin
 
 # Loads custom tasks from `lib/capistrano/tasks' if you have any defined.
 # tasks included: passenger, checksum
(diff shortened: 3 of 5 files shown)


config/deploy.rb
```diff
@@ -144,12 +145,12 @@
  after :symlink_shared, :remove_resque
 
  # Restart the application
- desc 'Restart application'
- task :restart do
-  on roles(:app), in: :sequence, wait: 5 do
-      execute :touch, release_path.join('tmp/restart.txt')
-  end
- end
+ #desc 'Restart application'
```
mjgiarlo: Feel free to delete these lines if we're not using them; we can always use git to pull them back later if need be.

db/schema.rb
```diff
@@ -190,8 +190,8 @@
     t.string   "label"
     t.string   "lowerLabel"
     t.string   "uri"
-    t.datetime "created_at", default: '2014-05-12 17:29:21'
-    t.datetime "updated_at", default: '2014-05-12 17:29:21'
+    t.datetime "created_at", default: '2014-10-28 15:10:55'
```
mjgiarlo: Do we need changes to the schema to support capistrano3?

lib/capistrano/submodule_strategy.rb
```diff
@@ -0,0 +1,32 @@
+class Capistrano::Git < Capistrano::SCM
+  module SubmoduleStrategy
+    include DefaultStrategy
+
+    # check for a .git directory
+  def test
+    test! " [ -d #{repo_path}/.git ] "
+  end
+ 
+  # same as in Capistrano::Git::DefaultStrategy
```
mjgiarlo: If this method is the same as one pulled in via the include, do we need it in here?

lib/capistrano/submodule_strategy.rb
```diff
@@ -0,0 +1,32 @@
...
+  def check
+    test! :git, :'ls-remote', repo_url
+  end
+ 
+  def clone
+    git :clone, '-b', fetch(:branch), '--recursive', repo_url, repo_path
+  end
+ 
+  # same as in Capistrano::Git::DefaultStrategy
```
mjgiarlo: If this method is the same as one pulled in via the include, do we need it in here?

lib/capistrano/submodule_strategy.rb
```diff
@@ -0,0 +1,32 @@
...
+  def update
+    git :remote, :update
+  end
+ 
+    def release
+      context.execute :rm, '-rf', release_path
```
mjgiarlo, line 14: What's this deleting?

mjgiarlo:
Do we need the empty .capignore file?

mjgiarlo:
Hey @pcrum!  I left a handful of comments, and Travis is currently showing failing tests.  Once you get tests passing, can you squash your commits?  Thanks for the PR.

config/deploy.rb
pcrum: Not a problem, will do

db/schema.rb
pcrum: No, no schema changes are needed. May have to review when this was added.

lib/capistrano/submodule_strategy.rb
pcrum: This module overrides the default git deploy strategy of capistrano 3 where the release path deployment is set using git archive where this overrides the method release method to remove the current release path and then clone the repo, branch, and any recursive submodules that may exist.

lib/capistrano/submodule_strategy.rb
pcrum: I think some of the methods are the same as the default but the important method override is the release method.

config/deploy.rb
mjgiarlo: Thx!  I love :bomb:ing code. :)

db/schema.rb
mjgiarlo: K.

lib/capistrano/submodule_strategy.rb
mjgiarlo: OK. So we should only need to define methods that differ from the implementations already provided by the module, and can remove the others marked as "same as?"

pcrum:
Hey @mjgiarlo , I believe I've corrected the formatting. Please take a look when you have a moment. Thanks

mjgiarlo:
Looks great, @pcrum!  I believe the Travis error above is unrelated to this work, so I'm inclined to merge this.  Thoughts, @cam156 @hectorcorrea @awead @mtribone?

hectorcorrea:
@mjgiarlo @pcrum I agree that the error does not seem to be related to the changes. I've restarted the build and will check again in 20 or so minutes.

mjgiarlo:
:clap:

awead:
Looks good to me, I'm pulling the trigger.

mjgiarlo:
:+1: @pcrum

PR #39, opened 2014-11-10
Make the regeneration of the sitemap.xml an scheduled job to be executed...

hectorcorrea:
... asynch rather than during the deployment


mjgiarlo:
:+1:

PR #42, opened 2014-12-01
Fixes size of marketing heading at 992px min, refs #9621

mtribone:
Small detail between media queries of certain sizes.


mjgiarlo:
:ok_hand:

PR #43, opened 2014-12-09
Integrated sufia's user stats.  Story #9305

val99erie:


carolyncole:
@val99erie I think this override (https://github.com/psu-stewardship/scholarsphere/blob/develop/app/models/file_usage.rb) needs to be removed (which was part of another ticket, but I did not do it yet) for your new configuration additions to make send.

And this line too:https://github.com/psu-stewardship/scholarsphere/blob/develop/config/application.rb#L99  Otherwise we are configuring  things in two places.

Unfortunately you and I are worked on similar things and you are getting done first.  Do you mind removing those two overrides too?

val99erie:
@cam156 I made those 2 changes

PR #61, opened 2015-01-27
Use hydra-jetty 8.1 unless ENV['MIGRATE'] is provided

jcoyne:


lib/tasks/jetty.rake
```diff
@@ -1,12 +1,14 @@
...
-namespace :jetty do
-  desc "Apply all Solr and Fedora configs to Jetty, including full-texting searching"
-  task :config do
-    Rake::Task["sufia:jetty:download_jars"].invoke
-    Rake::Task["hydra:jetty:config_solr"].invoke
+  namespace :jetty do
+    desc "Apply all Solr and Fedora configs to Jetty, including full-texting searching"
+    task :config do
+      Rake::Task["sufia:jetty:download_jars"].invoke
+      Rake::Task["hydra:jetty:config_solr"].invoke
```
mjgiarlo, line 11: Could you update/replace these two lines with:

``` ruby
Rake::Task['sufia:jetty:config'].invoke
```

jcoyne, line 11: @mjgiarlo nope, I just fixed the indenting. That sounds like a change of functionality which should be in its own PR.

PR #62, opened 2015-01-27
Fixes button size, input spacing, colors. refs #9732

mtribone:


mjgiarlo:
Are these ScholarSphere-specific changes, or could/should this go into Sufia?  Otherwise looks :ok_hand: to me!

mtribone:
These are ScholarSphere specific, but could be pushed to Sufia if the community wanted them. I had to make a quick fix to change my hex colors in the style sheet to a variable. Just rebased and force pushed.

PR #69, opened 2015-01-28
Abandon references to noids per the latest in Sufia

mjgiarlo:


lib/tasks/scholarsphere.rake
```diff
@@ -98,7 +98,7 @@ namespace :scholarsphere do
     #loop through each page appending the ids to the original list
     while idList.length < totalNum
        page += 1
-       resp = query_solr(q:"{!lucene q.op=AND df=text}id:#{ScholarSphere::Application.config.id_namespace}\\:* has_model_s:*GenericFile*", page:page)
+       resp = query_solr(q:"{!lucene q.op=AND df=text}id:#{ScholarSphere::Application.config.redis_namespace}\\:* has_model_s:*GenericFile*", page:page)
```
carolyncole, line 101: I do not think we need the namespace in the solr query anymore

lib/tasks/scholarsphere.rake
```diff
@@ -89,7 +89,7 @@ namespace :scholarsphere do
   desc "Characterize all files"
   task characterize: :environment do
     # grab the first increment of document ids from solr
-    resp = query_solr(q:"{!lucene q.op=AND df=text}id:#{ScholarSphere::Application.config.id_namespace}\\:* has_model_s:*GenericFile*" )
+    resp = query_solr(q:"{!lucene q.op=AND df=text}id:#{ScholarSphere::Application.config.redis_namespace}\\:* has_model_s:*GenericFile*" )
```
carolyncole, line 92: I do not think the namespace is needed in the query

spec/fixtures/scholarsphere_generic_stub.foxml.erb
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<foxml:digitalObject PID="<%=ScholarSphere::Application.config.id_namespace%>:<%= @id %>" VERSION="1.1" xmlns:foxml="info:fedora/fedora-system:def/foxml#"
```
carolyncole, line 2: I wonder if this foxml is still valid, and something we want.  I do not think we have used it much

lib/tasks/scholarsphere.rake
mjgiarlo, line 92: I agree!  I intentionally didn't change it because it's orthogonal to the PR (which, in this case, is just making sure things don't break because we reference an invalid config key).

lib/tasks/scholarsphere.rake
mjgiarlo, line 101: See above.

spec/fixtures/scholarsphere_generic_stub.foxml.erb
mjgiarlo, line 2: I wonder the same!  See above re: these changes. It might be worth adding two new tickets, one for fixing our rake tasks to not use the redis_namespace, and another for seeing if we still want these fixtures around.  Thx.

jcoyne:
:+1:

PR #70, opened 2015-01-28
Adding back in user notification when permissions are modified.  refs #9...

carolyncole:
...677

### DIFF
--- a/lib/devise/behaviors/permissions_notification_behavior.rb
+++ b/lib/devise/behaviors/permissions_notification_behavior.rb
@@ -11,41 +11,19 @@ module Behaviors
     # }
     def evaluate_permission_state(previous_permissions,current_permissions)
       state = {}
-      chash = {}
-      phash = {}
-
-      #converts current permissions to key value hash where user id is hash
-      chash = current_permissions.inject({}){|result,el| result[el[:name]]=el; result} unless current_permissions.nil?
-
-      #converts previous permissions to key value hash where user id is hash
-      phash = previous_permissions.inject({}){|result,el| result[el[:name]]=el; result} unless previous_permissions.nil?
 
       #append added permissions to the state hash
-      state[:added] = {}
-      (chash.keys - phash.keys).each{ |key| state[:added][key] = chash[key].deep_dup }
+      state[:added] = current_permissions.to_a - previous_permissions.to_a
 
       #append removed permissions to the state hash
-      state[:removed] = {}
-      (phash.keys - chash.keys).each{ |key| state[:removed][key] = phash[key].deep_dup }
-
-      #append modified or unchanged permissions to the state hash
-      state[:modified] = {}
-      state[:unchanged] = {}
-      (phash.keys & chash.keys).each do |key|
-        if phash[key].eql?(chash[key])
-          state[:unchanged][key] = phash[key].deep_dup
-        else
-          state[:modified][key] = phash[key].deep_dup
-          state[:modified][key][:previous_access] = chash[key][:access]
-        end
-      end
+      state[:removed] = previous_permissions.to_a - current_permissions.to_a
 
       #return state
       return state
     end
 
     # Overriding Sufia::FilesControllerBehavior to catch permission state broadcasting.
-    def notify_users(permission_state)
+    def notify_users(permission_state, generic_file)
       #retrieve batch user
       sender = User.batchuser
 
@@ -54,13 +32,13 @@ module Behaviors
 
         #iterate through added permissions, check that type is user,
         #locate and notify user of permissions change
-        permission_state[:added].each do |userid,entry|
-          if entry[:type]=="user"
-            recipient = User.find_by_user_key(userid)
+        permission_state[:added].each do |permission|
+          if permission[:type]=="person"
+            recipient = User.find_by_user_key(permission[:name])
             sender.send_message(recipient,
                                 t("scholarsphere.notifications.permissions.notification",
-                                  file: @generic_file.title,
-                                  access: entry[:access]),
+                                  file: generic_file.title,
+                                  access: permission[:access]),
                                 t("scholarsphere.notifications.permissions.subject")) unless recipient.nil?
 
           end
--- a/spec/lib/devise/behaviors/permission_notification_behavior_spec.rb
+++ b/spec/lib/devise/behaviors/permission_notification_behavior_spec.rb
@@ -1,0 +1,115 @@
+require 'spec_helper'
+require 'i18n'
+
+class NotifiableThing
+  include Behaviors::PermissionsNotificationBehavior
+
+  def evaluate(original_perm, new_perm)
+    evaluate_permission_state(original_perm, new_perm)
+  end
+
+  def notify(state, generic_file)
+    notify_users(state, generic_file)
+  end
+
+  def t(key, *args)
+    I18n.t(key, *args)
+  end
+end
+
+describe Behaviors::PermissionsNotificationBehavior do
+  subject { NotifiableThing.new }
+  describe "evaluate_permission_state" do
+    let(:no_perm) { [] }
+    let(:one_perm) { [{ name: "abd123", type: "person", access: "edit" }] }
+    let(:multi_perm) { [{ name: "zzz123", type: "person", access: "edit" },{ name: "def123", type: "person", access: "edit" }] }
+
+    context "no permissions after" do
+      it "has nothing added or removed for no permissions before" do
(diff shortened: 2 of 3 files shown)


lib/devise/behaviors/permissions_notification_behavior.rb
```diff
@@ -11,41 +11,19 @@ module PermissionsNotificationBehavior
...
-      state[:unchanged] = {}
-      (phash.keys & chash.keys).each do |key|
-        if phash[key].eql?(chash[key])
-          state[:unchanged][key] = phash[key].deep_dup
-        else
-          state[:modified][key] = phash[key].deep_dup
-          state[:modified][key][:previous_access] = chash[key][:access]
-        end
-      end
+      state[:removed] = previous_permissions.to_a - current_permissions.to_a
```
mjgiarlo, line 19: I see we no longer track `state[:modified]` anymore. Is that because all of those changes wind up recorded as additions and removals?

spec/lib/devise/behaviors/permission_notification_behavior_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+require 'i18n'
+
+class NotifiableThing
+  include Behaviors::PermissionsNotificationBehavior
+
+  def evaluate(original_perm, new_perm)
+    evaluate_permission_state(original_perm, new_perm)
+  end
+
+  def notify(state, generic_file)
```
mjgiarlo, line 11: Why redefine `notify_users` and `evaluate_permission_state`?

spec/lib/devise/behaviors/permission_notification_behavior_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+    notify_users(state, generic_file)
+  end
+
+  def t (key, *args)
```
mjgiarlo: Remove whitespace between method name and parenthesis.

mjgiarlo: And why are we defining the `t` method?  Trying to grok why the `NotifiableThing` class isn't just a one-liner with the `include` statement.

spec/lib/devise/behaviors/permission_notification_behavior_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+    I18n.t(key, *args)
+  end
+end
+
+describe Behaviors::PermissionsNotificationBehavior do
+  subject {NotifiableThing.new}
```
mjgiarlo: Add one space on the inside of each of the braces: `subject { NotifiableThing.new }`

spec/lib/devise/behaviors/permission_notification_behavior_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+  describe "evaluate_permission_state" do
+    let(:no_perm) { [] }
+    let(:one_perm) { [{:name=>"abd123", :type=>"person", :access=>"edit"}] }
```
mjgiarlo: Hash rockets baaaaaaaad.

spec/lib/devise/behaviors/permission_notification_behavior_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+    let(:multi_perm) { [{:name=>"zzz123", :type=>"person", :access=>"edit"},{:name=>"def123", :type=>"person", :access=>"edit"}] }
```
mjgiarlo: No hash rockets, _por favor_!

spec/lib/devise/behaviors/permission_notification_behavior_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+        expect(state[:removed]).to eq one_perm
+      end
+      it "has permissions removed for multiple permissions before" do
+        state = subject.evaluate(multi_perm, no_perm)
+        expect(state[:added]).to eq []
+        expect(state[:removed]).to eq multi_perm
+      end
+    end
+    context "one permission after" do
+      it "has nothing added or removed for one permission before" do
```
mjgiarlo, line 45: What if the one perm has view access and the other edit?  What if there's a difference in name only?  Or in name and type (removing edit access from a user, and granting it to a group)?  Did we lose that in the rewrite?

spec/lib/devise/behaviors/permission_notification_behavior_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+        state = subject.evaluate(one_perm, one_perm)
+        expect(state[:added]).to eq []
+        expect(state[:removed]).to eq []
+      end
+      it "has permission added for no permissions before" do
+        state = subject.evaluate(no_perm, one_perm)
+        expect(state[:added]).to eq one_perm
+        expect(state[:removed]).to eq []
+      end
+      it "has permissions added and removed for no permissions before" do
```
mjgiarlo: Description says "no permissions before" but it's multi perm before.

spec/lib/devise/behaviors/permission_notification_behavior_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+    context "permissions added" do
+      let (:state){ { added: [{:name=>"abd123", :type=>"person", :access=>"edit"}], removed: [] } }
+      it "notifies one user added" do
+        expect(batch_user).to receive(:send_message).with(user, message, message_subject)
+        subject.notify(state, generic_file)
+      end
+    end
+    context "permissions removed" do
+      let (:state){ { added: [ ], removed: [{:name=>"abd123", :type=>"person", :access=>"edit"}] } }
+      it "notifies no one" do
```
mjgiarlo, line 109: Why no notification?

jcoyne:
Why is this stuff in `lib/devise`?

spec/lib/devise/behaviors/permission_notification_behavior_spec.rb
carolyncole, line 11: They are protected so this was my cheat at being able to call them

spec/lib/devise/behaviors/permission_notification_behavior_spec.rb
carolyncole: see above...

spec/lib/devise/behaviors/permission_notification_behavior_spec.rb
carolyncole, line 45: There were absolutely no tests for anything, so we may or may not have lost things...
I was putting in some basic test, which are better than nothing

spec/lib/devise/behaviors/permission_notification_behavior_spec.rb
carolyncole, line 109: That is what the code originally did.  The original code sucks, but I was just updating it to work the same way it did.  We can add a ticket into the system to make it better if we want more notification.

carolyncole:
I have no idea.  The ticket was to add back in previous functionality, but I can move it...

lib/devise/behaviors/permissions_notification_behavior.rb
carolyncole, line 19: Yes, the old permission will be a removal and the new one will be an addition.  We were never using the modified or removed permissions

spec/lib/devise/behaviors/permission_notification_behavior_spec.rb
carolyncole, line 45: @mjgiarlo The only code only dealt with users, so group access was not considered.  This code was supposed to be a first step, but the Paul got pulled on to other tasks.  We should likely create a ticket or resurrect the old one, to work on this more later.

carolyncole, line 45: Here is the ticket.  It is still open... 
https://scm.dlt.psu.edu/issues/1502

lib/devise/behaviors/permissions_notification_behavior.rb
mjgiarlo, line 19: OK!

spec/lib/devise/behaviors/permission_notification_behavior_spec.rb
mjgiarlo, line 11: Ahhhh, :ok_hand:

spec/lib/devise/behaviors/permission_notification_behavior_spec.rb
mjgiarlo, line 45: Very good.

PR #73, opened 2015-01-30
Rake tasks to detect and create missing batches

hectorcorrea:
Addresses RedMine ticket 9740 
https://scm.dlt.psu.edu/issues/9740

### DIFF
--- a/lib/tasks/premigrate.rake
+++ b/lib/tasks/premigrate.rake
@@ -1,0 +1,102 @@
+require 'action_view'
+require 'blacklight/solr_helper'
+require 'rainbow'
+
+include ActionView::Helpers::NumberHelper
+include Blacklight::SolrHelper
+
+namespace :pre_migrate do
+
+  ALL_ROWS = 1_000_000
+
+  def logger
+    Rails.logger
+  end
+
+  def get_all_batches
+    batches = []
+    query = Solrizer.solr_name('has_model', :symbol) + ':"' + Batch.to_class_uri + '"'
+    results = ActiveFedora::SolrService.query(query, fl: "id", rows: ALL_ROWS)
+    results.each do |batch|
+      batch_id = batch["id"]
+      if valid_id?(batch_id) && !batches.include?(batch_id)
+        batches.push batch_id 
+      end
+    end
+    batches
+  end
+
+  # batch_token should be an array of one element
+  # ["info:fedora/scholarsphere:g732d900g"]
+  def get_batch_id(batch_token)
+    return if batch_token.blank?
+    batch_id = batch_token.first.split('/').last
+    batch_id
+  end
+
+  def valid_id?(identifier)
+    identifier.to_s.start_with?(ScholarSphere::Application.config.id_namespace)
+  end
+
+  def get_batches_in_files
+    batches = []
+    query = Solrizer.solr_name('has_model', :symbol) + ':"' + GenericFile.to_class_uri + '"' 
+    results = ActiveFedora::SolrService.query(query, fl: "id is_part_of_ssim", rows: ALL_ROWS)
+    results.each do |file|
+      begin
+        batch_id = get_batch_id file['is_part_of_ssim'] 
+        if valid_id?(batch_id) && !batches.include?(batch_id) 
+          batches.push batch_id   
+        end
+      rescue Exception => e  
+        logger.error "Error getting batch from file #{file}"
+      end
+    end
+    batches
+  end
+
+  def get_missing_batches(verbose)
+    all_batches = get_all_batches
+    batches_in_files = get_batches_in_files
+    missing_batches = []
+    batches_in_files.each do |id|
+      unless all_batches.include? id
+        missing_batches.push(id)
+      end
+    end
+    if verbose
+      logger.info "Total batches: #{all_batches.length}"
+      logger.info all_batches
+      logger.info "Batches in files: #{batches_in_files.length}"
+      logger.info batches_in_files
+    end
+    missing_batches
+  end
+
+  desc "Get a list of missing batches"
+  task "list_missing_batches", [:verbose] => :environment do |cmd, args|
+    verbose = args[:verbose] == "true"
+    missing_batches =  get_missing_batches(verbose)
+    logger.info "Missing batches: #{missing_batches.length}"
+    logger.info missing_batches
+  end
+
+  desc "Create missing batches"
+  task create_missing_batches: :environment do
+    missing_batches = get_missing_batches(false)
+    if missing_batches.empty?
+      logger.info "No missing batches need to be created"
+    else
+      logger.info "Creating #{missing_batches.length} batches..."
+      missing_batches.each do |pid|
+        begin
+          logger.info "Batch: #{pid}"
+          Batch.create(pid: pid)
+        rescue Exception => e  
+          logger.error "\tError creating batch #{pid}: #{e.message} #{e.backtrace.inspect}"          
+        end
+      end
+    end
+  end
+end
+


lib/tasks/premigrate.rake
```diff
@@ -0,0 +1,109 @@
+require 'action_view'
+require 'blacklight/solr_helper'
+require 'rainbow'
+
+include ActionView::Helpers::NumberHelper
+include Blacklight::SolrHelper
+
+namespace :pre_migrate do
+
+
```
mjgiarlo: Can you condense all the double blank lines in this file to single blank lines?

lib/tasks/premigrate.rake
```diff
@@ -0,0 +1,109 @@
...
+  ALL_ROWS = 1_000_000
```
mjgiarlo, line 10: Nice, I like the underscore notation here for readability!

lib/tasks/premigrate.rake
```diff
@@ -0,0 +1,109 @@
...
+
+
+  def logger
+    Rails.logger
+  end
+
+
+  def get_all_batches
+    batches = []
+    query = Solrizer.solr_name('has_model', :symbol) + ':"' + Batch.to_class_uri + '"'
```
mjgiarlo, line 18: I'm on the fence about this, but what do you think about:

``` ruby
query = "#{Solrizer.solr_name('has_model', :symbol)}:\"#{Batch.to_class_uri}\""
```

? Less readable or more readable?  On one hand, I prefer the lack of pluses, but on the other I find the escaped double-quotes fugly.

lib/tasks/premigrate.rake
```diff
@@ -0,0 +1,109 @@
...
+    results = ActiveFedora::SolrService.query(query, :fl=>"id", :rows => ALL_ROWS)
```
mjgiarlo: Nuke the hashrockets!

``` ruby
results = ActiveFedora::SolrService.query(query, fl: 'id', rows: ALL_ROWS)
```

lib/tasks/premigrate.rake
```diff
@@ -0,0 +1,109 @@
...
+    results.each do |batch|
+      batch_id = batch["id"]
+      is_valid_id = batch_id != nil && batch_id.start_with?(ScholarSphere::Application.config.id_namespace)
```
mjgiarlo: Change `id_namespace` to `redis_namespace`, since the former was renamed in Sufia.

mjgiarlo: A minor point, but I find that predicate variables in Ruby (usually) tend not to have verb prefixes.  So I'd replace with:

``` ruby
valid_id = !batch_id.nil? && batch_id.start_with?(ScholarSphere::Application.config.redis_namespace)
```

Or, alternatively, coerce to string rather than check for nil ("tell, don't ask"), and move this into its own method:

``` ruby
def valid_id?(identifier)
  identifier.to_s.start_with?(ScholarSphere::Application.config.redis_namespace)
end
```

lib/tasks/premigrate.rake
```diff
@@ -0,0 +1,109 @@
...
+      end
+    end
+    batches
+  end
+
+
+  # batch_token should be an array of one element
+  # ["info:fedora/scholarsphere:g732d900g"]
+  def get_batch_id(batch_token)
+    return nil if batch_token == nil || batch_token.length == 0
```
mjgiarlo: You can return without explicit nil here, and can also use nil?, like so:

``` ruby
return if batch_token.nil? || batch_token.empty?
```

Or if the blank method is available in this rake task, even simpler:

``` ruby
return if batch_token.blank?
```

Which will return true if batch_token is nil, an empty array, or an empty string.

lib/tasks/premigrate.rake
```diff
@@ -0,0 +1,109 @@
...
+    batch_id = batch_token[0].split('/').last
+    return nil if !batch_id.start_with?(ScholarSphere::Application.config.id_namespace)
```
mjgiarlo: If you define `valid_id?` per the above comment, you can replace this line with:

``` ruby
return unless valid_id?(batch_id)
```

lib/tasks/premigrate.rake
```diff
@@ -0,0 +1,109 @@
...
+    query = Solrizer.solr_name('has_model', :symbol) + ':"' + GenericFile.to_class_uri + '"' 
+    results = ActiveFedora::SolrService.query(query, :fl=>"id is_part_of_ssim", :rows => ALL_ROWS)
+    results.each do |file|
+      begin
+        batch_id = get_batch_id file['is_part_of_ssim'] 
+        unless batch_id == nil || batches.include?(batch_id) 
+          batches.push batch_id   
+        end
+      rescue Exception => e  
+        puts "Error getting batch from file #{file}"
```
mjgiarlo: Use logger instead of puts.

lib/tasks/premigrate.rake
```diff
@@ -0,0 +1,109 @@
...
+  # ["info:fedora/scholarsphere:g732d900g"]
+  def get_batch_id(batch_token)
+    return nil if batch_token == nil || batch_token.length == 0
+    batch_id = batch_token[0].split('/').last
+    return nil if !batch_id.start_with?(ScholarSphere::Application.config.id_namespace)
+    batch_id
+  end
+
+
+  def get_batches_in_files
```
mjgiarlo, line 41: Some of the comments above also apply to this method.

lib/tasks/premigrate.rake
```diff
@@ -0,0 +1,109 @@
...
+      puts batches_in_files
+    end
+    missing_batches
+  end
+
+
+  desc "Get a list of missing batches"
+  task "list_missing_batches", [:verbose] => :environment do |cmd, args|
+    verbose = args[:verbose] == "true"
+    puts "Solr at #{ActiveFedora.solr_config[:url]}"
```
mjgiarlo: Debug leftover?

lib/tasks/premigrate.rake
```diff
@@ -0,0 +1,109 @@
...
+    missing_batches =  get_missing_batches(verbose)
+    puts "Missing batches: #{missing_batches.length}"
```
mjgiarlo: Use logger instead of puts?

lib/tasks/premigrate.rake
```diff
@@ -0,0 +1,109 @@
...
+    puts missing_batches
+  end
+
+
+  desc "Create missing batches"
+  task create_missing_batches: :environment do
+    missing_batches = get_missing_batches(false)
+    if missing_batches.length == 0
```
mjgiarlo: Use `empty?` here.

lib/tasks/premigrate.rake
```diff
@@ -0,0 +1,109 @@
...
+      puts "No missing batches need to be created"
```
mjgiarlo: Use logger here.

lib/tasks/premigrate.rake
```diff
@@ -0,0 +1,109 @@
...
+    else
+      puts "Creating #{missing_batches.length} batches..."
```
mjgiarlo: Logger, logger, logger. I'm getting thirsty.

lib/tasks/premigrate.rake
hectorcorrea: Done

lib/tasks/premigrate.rake
hectorcorrea, line 18: I tried with string interpolation and I thought it looked more confusing. I think the pluses and the spacing make the (tricky) expression to stand out in this case.

lib/tasks/premigrate.rake
hectorcorrea: :bomb: ed the hashrockets

lib/tasks/premigrate.rake
hectorcorrea: Cool idea on the valid_id function. 

I have to use `id_namespace` because this will run in ScholarSphere with the Fedora 3 stack.

lib/tasks/premigrate.rake
hectorcorrea: I keep forgetting about Rail's `blank?`. Fixed.

lib/tasks/premigrate.rake
hectorcorrea: I ended up moving the validation out of this method so that the method does only one thing (gets the id) and left the validation (now in the function that you suggested) as part of the method that calls it which had the extra benefit of make the logic clearer.

lib/tasks/premigrate.rake
hectorcorrea: Done

lib/tasks/premigrate.rake
hectorcorrea, line 41: Thanks.

lib/tasks/premigrate.rake
hectorcorrea: Yup.

lib/tasks/premigrate.rake
hectorcorrea: Done.

lib/tasks/premigrate.rake
hectorcorrea: Done.

lib/tasks/premigrate.rake
hectorcorrea: Done.

lib/tasks/premigrate.rake
hectorcorrea: Done, done, done :)

hectorcorrea:
I will re-submit this PR with a different target branch.

PR #76, opened 2015-02-04
Rake taks to gather information about all objects in our Fedora 3 repo

hectorcorrea:
Rake taks to gather information about all objects in our Fedora 3 repo to audit our data before and after a migration.

Running `bundle exec rake pre_migrate:f3_audit` will dump to a SQL table the list of all objects (in our namespace) in our Fedora 3 repo. The table will include the PID, the model (e.g. GenericFile, Batch), and the title of each object.

This code does not use neither Rubydora nor ActiveRecord. It issues straight HTTP calls to the Fedora API to fetch the information (although I stole from Rubydora's Search implementation how to get the pagination to work.) This code does NOT fetch the content of the objects, it only fetches the XML with their metadata. 

There are a couple of rake tasks that only query the Fedora 3 repo and dump the information to the console if you want just to test the code. `bundle exec rake pre_migrate:get_pids` gives you a list of PIDS whereas `bundle exec rake pre_migrate:get_info` give you a list of PIDs with their model and title.


hectorcorrea:
Could somebody merge this PR? (Notice that it targets a branch still running on Fedora 3 since it contains Rake tasks to run against the Fedora 3 repo to gather some basic information about the objects that will be migrated to Fedora 4)

mjgiarlo:
@cam156 @awead @mtribone has anyone had a chance to review this yet?  I haven't.  Sorry for the delay, @hectorcorrea

hectorcorrea:
@mjgiarlo no. I thought this was the one that you already had looked at but that was another rake task. But since you are here...could you give it a look and merge it, err, provide some feedback.

db/schema.rb
```diff
@@ -40,10 +40,9 @@
     t.text     "value"
     t.datetime "created_at"
     t.datetime "updated_at"
+    t.string   "external_key"
   end
 
-  add_index "content_blocks", ["name"], name: "index_content_blocks_on_name", unique: true, using: :btree
-
```
awead, line 46: Is this related to your tables?

db/schema.rb
```diff
@@ -40,10 +40,9 @@
     t.text     "value"
     t.datetime "created_at"
     t.datetime "updated_at"
+    t.string   "external_key"
```
awead, line 43: This also seems unrelated to adding a new table. Maybe Rails is doing something new here?

db/schema.rb
```diff
@@ -214,8 +222,8 @@
     t.string   "label"
     t.string   "lowerLabel"
     t.string   "uri"
-    t.datetime "created_at", default: '2014-12-10 16:14:48'
-    t.datetime "updated_at", default: '2014-12-10 16:14:48'
+    t.datetime "created_at", default: '2014-08-20 15:18:33'
+    t.datetime "updated_at", default: '2014-08-20 15:18:33'
```
awead: I don't think the date values matter, but we should be sure.

app/models/migrate_audit.rb
```diff
@@ -0,0 +1,15 @@
+class MigrateAudit < ActiveRecord::Base
+
+  def self.f3_audit(fedora_url, fedora_user, fedora_password, namespace)
```
awead, line 5: You could get all these values from config files, I believe.

app/models/migrate_audit_fedora3.rb
```diff
@@ -0,0 +1,126 @@
+class MigrateAuditFedora3
+
+  def initialize(fedora_url, fedora_user, fedora_password, namespace)
```
awead, line 3: Same as above comment. You could get these values from config files.

mtribone:
I'd be more than happy to click the shiny, green button  :-)

app/models/migrate_audit_fedora3.rb
hectorcorrea, line 3: @awead I tried to keep this class (and the Model) as free of dependencies as possible and that's why I am passing these values as parameters. In the rake task that instantiate these classes I am indeed reading these parameters from the config as you suggested.

awead:
@mjgiarlo @cam156 I've made some notes. The approach looks sound to me. There are no tests, however, and while I know @hectorcorrea has been testing it against our production data, it might be a good idea to have a few integration tests that use Rubydora to verify the counts.

db/schema.rb
hectorcorrea: @awead I'll take a closer look into the changes to schema.rb -- all of them were automatically added by Rails when I ran the generator to create the new `migrate_audits` table but the diff makes it look like it changed other tables which I am not sure is correct.

PR #78, opened 2015-02-20
Adds rake tasks to audit the Fedora 3 to Fedora 4 migration

hectorcorrea:
This pull request adds a few rake tasks to audit the migration of objects from Fedora 3 to Fedora 4. The idea is that these tasks will be run after deploying the Fedora 4 code base to production and immediately after the migration of data has been performed.

The code to perform the audit issues straight HTTP requests to Fedora 3 and Fedora 4 and bypasses Rubydora and ActiveFedora to make sure we query the Fedora repo as close to the metal as possible. 

The following Wiki page provides some information on how these rake tasks will be run in production: https://github.com/psu-stewardship/scholarsphere/wiki/Fedora-3-to-Fedora-4-Migration

### DIFF
--- a/app/models/migrate_audit.rb
+++ b/app/models/migrate_audit.rb
@@ -1,0 +1,35 @@
+class MigrateAudit < ActiveRecord::Base
+
+  # This method populates the migrate_audit table with data 
+  # from the Fedora 3 repo.
+  def self.f3_audit!(auditor)
+    raise ArgumentError, "Auditor must respond to audit method" unless auditor.respond_to? :audit
+    clear_audit_data!
+    auditor.audit do |f3_obj|
+      create(f3_pid: f3_obj.pid, f3_model: f3_obj.has_model, f3_title: f3_obj.title)
+    end
+  end
+
+  # This method makes sure each record in the migrate_audit table is
+  # in the Fedora 4 repo and it has the same model as its Fedora 3
+  # counterpart.
+  def self.f4_audit(auditor)
+    raise ArgumentError, "Auditor must respond to audit method" unless auditor.respond_to? :audit
+    reset_audit_status
+    auditor.audit(all) do |result|
+      MigrateAudit.update(result.id, f4_id: result.f4_id, status: result.status)
+    end
+  end
+
+  private 
+
+  def self.clear_audit_data!
+    # destroy all existing audit records
+    all.each { |obj| obj.destroy! }
+  end
+
+  def self.reset_audit_status
+    MigrateAudit.update_all(f4_id: nil, status: nil)
+  end
+
+end
--- a/app/models/migrate_audit_fedora3.rb
+++ b/app/models/migrate_audit_fedora3.rb
@@ -1,0 +1,131 @@
+class MigrateAuditFedora3
+
+  Fedora3Object = Struct.new("Fedora3Object", :pid, :has_model, :title)
+
+  # The code in this class makes direct HTTP calls to Fedora 3 to 
+  # fetch data and does not depend on Rubydora or ActiveFedora.
+  def initialize(fedora_url, fedora_user, fedora_password, namespace)
+    @fedora_url = fedora_url
+    @fedora_user = fedora_user
+    @fedora_password = fedora_password
+    @namespace = namespace
+  end
+
+  def audit
+    raise "Must receive a block parameter" unless block_given?
+    pids.each do |pid|
+      f3_obj = get_info pid
+      yield f3_obj
+    end 
+  end
+
+  # Gets the list of all PIDs in the Fedora 3 repository. 
+  # Fedora returns the list of PIDS in an XML that looks more or
+  # less like this:
+  #
+  # <result>
+  #   <listSession>
+  #     <token>xxx</token>
+  #     ...
+  #   </listSession>
+  #   <resultList>
+  #     <objectFields>
+  #       <pid>scholarsphere:123xyz</pid>
+  #     <objectFields>
+  #     more objectFields
+  #   </resultList>
+  # <result>
+  def pids
+    batch_size = 50
+    all_pids = []
+    session_token = nil
+    while true
+      session = (session_token == nil) ? "" : "sessionToken=#{session_token}"
+      query_string = "query=&resultFormat=xml&maxResults=#{batch_size}&pid=true&#{session}"
+      fedora_response = fedora_get("/objects", query_string)
+      xml = Nokogiri::XML(fedora_response)
+      extract_pids(xml) do |pid| 
+        all_pids.push pid 
+      end
+      session_token = extract_session_token(xml)
+      break if session_token.nil?
+    end
+    all_pids  
+  end
+
+  # Gets basic information (model and title) about a Fedora 3 object.
+  # Fedora returns an XML that looks more ot less like this with the 
+  # information about a given object:
+  # 
+  # <foxml:digitalObject>
+  #   ...
+  #   <foxml:datastream ID="DC" ...>
+  #     <foxml:datastreamVersion ...>
+  #       <foxml:xmlContent>
+  #         <oai_dc:dc xmlns:dc="http://purl.org/dc/elements/1.1/" ...>
+  #           <dc:title>the_title_of_the_generic_file</dc:title>
+  #         </oai_dc:dc>
+  #       </foxml:xmlContent>
+  #     </foxml:datastreamVersion>
+  #   </foxml:datastream>
+  #   ...
+  #   <foxml:datastream ID="RELS-EXT" ...>
+  #     <foxml:datastreamVersion ...>
+  #       <foxml:xmlContent>
+  #         <rdf:RDF xmlns:ns1="info:fedora/fedora-system:def/model#" ...>
+  #           <rdf:Description rdf:about="info:fedora/scholarsphere:hh63sv96j">
+  #             <ns1:hasModel rdf:resource="info:fedora/afmodel:GenericFile"/>
+  #           </rdf:Description>
+  #         </rdf:RDF>
+  #       </foxml:xmlContent>
+  #     </foxml:datastreamVersion>
+  #   </foxml:datastream>
+  #   ...
+  def get_info(pid)
+    fedora_response = fedora_get("/objects/#{pid}/objectXML")
+    xml = Nokogiri::XML(fedora_response)
+    title = extract_title(xml)
+    has_model = extract_model(xml)
+    Fedora3Object.new(pid, has_model, title)
+  end  
+
+
+  private
+
+  def fedora_get(url_path, url_querystring = "")
+    uri = URI.parse(@fedora_url + url_path + "?" + url_querystring)
+    response = Net::HTTP.start(uri.hostname, uri.port) do |http|
+      request = Net::HTTP::Get.new(uri.path + "?" + url_querystring)
+      request.basic_auth(@fedora_user, @fedora_password)
+      http.request(request)
+    end
+    response.body
+  end
+
+  def extract_pids(xml)
+    docs = xml.xpath("//doc:pid", doc: "http://www.fedora.info/definitions/1/0/types/").children
+    docs.each do |doc|
+      yield doc.text if doc.to_s.start_with?(@namespace + ":")
+    end
+  end
+
+  # Get the session token. This is what allows us to fetch the next batch 
+  # of PIDs in the next execution of the query.
+  def extract_session_token(xml)
+    header = xml.xpath("//header:token", header: "http://www.fedora.info/definitions/1/0/types/").children
+    return header[0].text if header.length == 1
+  end
+
+  # Extract the title from the DC data stream
+  def extract_title(xml)
+    titles = xml.xpath("//dc:title", dc: "http://purl.org/dc/elements/1.1/").first
+    return titles.children[0].text if titles && titles.children.count > 0
+  end
+
+  # Extract the "has_model" (GenericFile, Batch, Collection) from the RELS-EXT data stream
+  def extract_model(xml)
+    models = xml.xpath("//ns1:hasModel", ns1: "info:fedora/fedora-system:def/model#" ).first
+    return models.attributes["resource"].value if models && models.attributes.key?("resource")
+  end
+
+end
--- a/app/models/migrate_audit_fedora4.rb
+++ b/app/models/migrate_audit_fedora4.rb
@@ -1,0 +1,92 @@
+class MigrateAuditFedora4
+
+  AuditResult = Struct.new("AuditResult", :id, :f3_pid, :f4_id, :status)
+
+  def initialize(fedora_url, fedora_user, fedora_password)
+    @fedora_url = fedora_url
+    @fedora_user = fedora_user
+    @fedora_password = fedora_password
+  end
+
+  def audit(f3_objects)
+    f3_objects.each do |f3_obj|
+      f4_url = @fedora_url + "/" + pid_to_uri(f3_obj.f3_pid)
+      result = AuditResult.new(f3_obj.id, f3_obj.f3_pid, f4_url, nil)
+      result.status = audit_one(f3_obj, f4_url)
+      yield result
+    end
+  end
+
+  private
+
+  def audit_one(f3_obj, f4_url)
+    f4_obj = fedora_get_metadata f4_url
+    return "Not found" if f4_obj.nil?
+    f4_model = model_from_resource f4_obj
+    f3_model = normalized_f3_model f3_obj.f3_model
+    return "Models mismatch. Expected: #{f3_model} but found: #{f4_model}" if f3_model != f4_model 
+    "OK"
+  end
+
+  # f3 models are stored like this: "info:fedora/afmodel:Batch"
+  # and we only care about the last part ("Batch" in this case)
+  def normalized_f3_model(f3_model)
+    f3_model.split(':').last
+  end
+
+  def pid_to_uri(pid)
+    id = pid.split(':').last
+    raise "Cannot detect ID from PID #{pid}" unless id.length >= 9
+    id[0..1] + "/" + id[2..3] + "/" + id[4..5] + "/" + id[6..7] + "/" + id
+  end
+
+  def model_from_resource(triples)
+    object = parse_triples triples 
+    object[:model]
+  end
+
+  # Response is an RDF graph in n-triple format and we expect it to have the 
+  # model of the object (e.g. GenericFile, Batch) as well a list of its children.
+  # 
+  # The model comes in the form:
+  #     <URI> <info:fedora/fedora-system:def/model#hasModel> "GenericFile"^^<http://www.w3.org/2001/XMLSchema#string> .
+  #
+  # Children come in the form:
+  #     <URI> <http://www.w3.org/ns/ldp#contains> <CHILD-URI-1> .
+  #     <URI> <http://www.w3.org/ns/ldp#contains> <CHILD-URI-2> .
+  #
+  def parse_triples response
+    model = nil
+    children = []
+    response.split("\n").each do |line|
+      tokens = line.split(" ")
+      predicate = tokens[1]
+      object = tokens[2]
+      if predicate == "<info:fedora/fedora-system:def/model#hasModel>"
+        # get the model of the object
+        has_model = tokens[2]
+        caret = object.index("^^")
+        if caret
+          model = has_model[0,caret].gsub('"', '')
+        end
+      elsif predicate == "<http://www.w3.org/ns/ldp#contains>"
+        # get the list of children of the object
+        child_url = tokens[2].gsub("<", "").gsub(">", "")
+        children.push child_url
+      end
+    end
+    {model: model, children: children}
+  end
+
+  def fedora_get_metadata(url)
+    headers = { "Accept" => "application/n-triples" }
+    uri = URI.parse(url  + "/fcr:metadata")
+    response = Net::HTTP.start(uri.hostname, uri.port) do |http|
+      request = Net::HTTP::Get.new(uri.path, headers)
+      request.basic_auth(@fedora_user, @fedora_password)
+      http.request(request)
+    end
+    return nil if response.code == "404"
+    response.body
+  end
+end
--- a/lib/tasks/premigrate.rake
+++ b/lib/tasks/premigrate.rake
@@ -5,7 +5,7 @@ require 'rainbow'
 include ActionView::Helpers::NumberHelper
 include Blacklight::SolrHelper
 
-namespace :pre_migrate do
+namespace :premigrate do
 
   ALL_ROWS = 1_000_000
 
@@ -73,6 +73,11 @@ namespace :pre_migrate do
     missing_batches
   end
 
+  def fedora3_settings(file_name)
+    fedora3_settings = YAML.load_file(file_name)
+    fedora3_settings[Rails.env].symbolize_keys
+  end
+
   desc "Get a list of missing batches"
   task "list_missing_batches", [:verbose] => :environment do |cmd, args|
     verbose = args[:verbose] == "true"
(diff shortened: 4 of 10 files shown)


hectorcorrea:
@awead @cam156 @mjgiarlo This is the latest on the code to audit the Fedora 3 to Fedora 4 migration. 

This code includes the suggestions that Adam made earlier this week (on a PR that I deleted.) Notice that I only added unit tests to test the code from the Fedora 4 point of view since I didn't want to be running tests against both Fedora 3 and Fedora 4 in the same code base. 

The unit tests stub the results that the Fedora 3 repo will have and then creates them in Fedora 4 and compares the two.

app/models/migrate_audit.rb
```diff
@@ -0,0 +1,30 @@
+class MigrateAudit < ActiveRecord::Base
+
+  # This method populates the migrate_audit table with data 
+  # from the Fedora 3 repo.
+  def self.f3_audit!(fedora_url, fedora_user, fedora_password, namespace)
```
awead: If looks like you're using these four arguments just to create `MigrateAuditFedora3`. Why not:

```
f3_audit!(MigrateAuditFedora3.new(fedora_url, fedora_user, fedora_password, namespace))
```

where `f3_audit!` is:

``` ruby
def self.f3_audit! auditor = nil
  rails StandardError unless auditor.kind_of?(MigrateAuditFedora3)
  ...
```

app/models/migrate_audit.rb
```diff
@@ -0,0 +1,30 @@
...
+    # destroy all existing audit records first
+    all.each { |obj| obj.destroy! }
+
+    audit = MigrateAuditFedora3.new(fedora_url, fedora_user, fedora_password, namespace)
+    f3_pids = audit.get_pids()
+    f3_pids.each do |pid|
```
awead: consider:
`audit.get_pids.each do |pid|`

app/models/migrate_audit.rb
```diff
@@ -0,0 +1,30 @@
...
+      info = audit.get_info(pid)
+      create(f3_pid: pid, f3_model: info[:has_model], f3_title: info[:title])
+    end
+  end
+
+  # This method makes sure each record in the migrate_audit table is
+  # in the Fedora 4 repo and it has the same model as its Fedora 3
+  # counterpart.
+  def self.f4_audit(fedora_url, fedora_user, fedora_password)
```
awead: see my previous comment. The args are only relevant to a MigrateAuditFedora4 object.

app/models/migrate_audit.rb
```diff
@@ -0,0 +1,30 @@
...
+    # Clear previous results
+    MigrateAudit.update_all(f4_id: nil, status: nil)
+    # Run the audit.
+    f3_data = all
+    audit = MigrateAuditFedora4.new(fedora_url, fedora_user, fedora_password)
+    audit.audit(f3_data) do |result|
```
awead: consider:
`audit.audit(all) do |result|`
and if you just passed a MigrateAuditFedora4 object, this line could read:
`auditor.audit(all) do |result|`
which reads a bit better.

app/models/migrate_audit_fedora3.rb
```diff
@@ -0,0 +1,130 @@
...
+  # fetch data and does not depend on Rubydora or ActiveFedora.
+  def initialize(fedora_url, fedora_user, fedora_password, namespace)
+    @fedora_url = fedora_url
+    @fedora_user = fedora_user
+    @fedora_password = fedora_password
+    @namespace = namespace
+  end
+
+  # Get the list of all PIDs in the Fedora 3 repository. 
+  def get_pids()
```
awead: `()` are unnecessary in Ruby.

app/models/migrate_audit_fedora3.rb
```diff
@@ -0,0 +1,130 @@
...
+      #     <token>xxx</token>
+      #     ...
+      #   </listSession>
+      #   <resultList>
+      #     <objectFields>
+      #       <pid>scholarsphere:123xyz</pid>
+      #     <objectFields>
+      #     more objectFields
+      #   </resultList>
+      # <result>
```
awead: This is documentation code and style-wise, it should be above the method call. See https://github.com/rdoc/rdoc for style pointers.

app/models/migrate_audit_fedora3.rb
```diff
@@ -0,0 +1,130 @@
...
+  def initialize(fedora_url, fedora_user, fedora_password, namespace)
+    @fedora_url = fedora_url
+    @fedora_user = fedora_user
+    @fedora_password = fedora_password
+    @namespace = namespace
+  end
+
+  # Get the list of all PIDs in the Fedora 3 repository. 
+  def get_pids()
+    batch_size = 50
```
awead, line 39: Does batch size always have to be 50? Why not make it an `attr_accessor` for flexibility.

app/models/migrate_audit_fedora3.rb
```diff
@@ -0,0 +1,130 @@
...
+    all_pids = []
+    session_token = nil
+    while true
+      session = (session_token == nil) ? "" : "sessionToken=#{session_token}"
+      query_string = "query=&resultFormat=xml&maxResults=#{batch_size}&pid=true&#{session}"
```
awead, line 44: Make `query_string` a method and you can simplify the session and session_token logic.

```
def query_string
  if session_token.nil?
    "query=&resultFormat=xml&maxResults=#{batch_size}&pid=true"
  else
    "query=&resultFormat=xml&maxResults=#{batch_size}&pid=true&#{session_token}"
  end
end
```

make an attr_accessor for session token and then `extract_session` is called directly to set or unset the session_token value.

app/models/migrate_audit_fedora3.rb
```diff
@@ -0,0 +1,130 @@
...
+      #   </listSession>
+      #   <resultList>
+      #     <objectFields>
+      #       <pid>scholarsphere:123xyz</pid>
+      #     <objectFields>
+      #     more objectFields
+      #   </resultList>
+      # <result>
+      fedora_response = fedora_get("/objects", query_string)
+      xml = Nokogiri::XML(fedora_response)
```
awead, line 46: why not:
`xml = Nokogiri::XML(fedora_get("/objects", query_string))`

app/models/migrate_audit_fedora3.rb
```diff
@@ -0,0 +1,130 @@
...
+      extract_pids(xml) do |pid| 
+        all_pids.push pid 
+      end
+      session_token = extract_session_token(xml)
+      break if session_token == nil
```
awead: break if session_token.nil?

app/models/migrate_audit_fedora3.rb
```diff
@@ -0,0 +1,130 @@
...
+    #       <foxml:xmlContent>
+    #         <rdf:RDF xmlns:ns1="info:fedora/fedora-system:def/model#" ...>
+    #           <rdf:Description rdf:about="info:fedora/scholarsphere:hh63sv96j">
+    #             <ns1:hasModel rdf:resource="info:fedora/afmodel:GenericFile"/>
+    #           </rdf:Description>
+    #         </rdf:RDF>
+    #       </foxml:xmlContent>
+    #     </foxml:datastreamVersion>
+    #   </foxml:datastream>
+    #   ...
```
awead: put documentation above the method call.

app/models/migrate_audit_fedora3.rb
```diff
@@ -0,0 +1,130 @@
...
+    fedora_response = fedora_get("/objects/#{pid}/objectXML")
+    xml = Nokogiri::XML(fedora_response)
+    title = extract_title(xml)
+    has_model = extract_model(xml)
+    {pid: pid, has_model: has_model, title: title}
```
awead: could condense this to:

```
xml = Nokogiri::XML(fedora_get("/objects/#{pid}/objectXML"))
{pid: pid, has_model: extract_model(xml), title: extract_title(xml)}
```

app/models/migrate_audit_fedora3.rb
```diff
@@ -0,0 +1,130 @@
...
+  end  
+
+
+  private
+
+  def fedora_get(url_path, url_querystring = "")
+    uri = URI.parse(@fedora_url + url_path + "?" + url_querystring)
+    response = Net::HTTP.start(uri.hostname, uri.port) { |http|
+      request = Net::HTTP::Get.new(uri.path + "?" + url_querystring)
```
awead, line 98: Not sure what the Ruby convention is, but I've seen `{}` for one-liners and `do ... end` for multiline stuff. @mjgiarlo any comment on that?

app/models/migrate_audit_fedora3.rb
```diff
@@ -0,0 +1,130 @@
...
+    title = extract_title(xml)
+    has_model = extract_model(xml)
+    {pid: pid, has_model: has_model, title: title}
+  end  
+
+
+  private
+
+  def fedora_get(url_path, url_querystring = "")
+    uri = URI.parse(@fedora_url + url_path + "?" + url_querystring)
```
awead, line 96: consider using `#{}` syntax for string interpolation as you did above.

app/models/migrate_audit_fedora3.rb
```diff
@@ -0,0 +1,130 @@
...
+      request = Net::HTTP::Get.new(uri.path + "?" + url_querystring)
+      request.basic_auth(@fedora_user, @fedora_password)
+      http.request(request)
+    }    
+    response.body
+  end
+
+  def extract_pids(xml)
+    docs = xml.xpath("//doc:pid", "doc" => "http://www.fedora.info/definitions/1/0/types/").children
+    docs.each do |doc|
```
awead, line 107: consolidate to `children.each`

app/models/migrate_audit_fedora3.rb
```diff
@@ -0,0 +1,130 @@
...
+      if doc.to_s.start_with?(@namespace + ":")
```
awead: I think starting with the namespace is enough. You don't need the colon... maybe?

app/models/migrate_audit_fedora3.rb
```diff
@@ -0,0 +1,130 @@
...
+        yield doc.text
```
awead: consolidate to:
`yield doc.text if doc.to_s.start_with ...`

app/models/migrate_audit_fedora3.rb
```diff
@@ -0,0 +1,130 @@
...
+      end
+    end
+  end
+
+  def extract_session_token(xml)
+    # Get the session token. This is what allows us to fetch the next batch 
+    # of PIDs in the next execution of the query.
+    header = xml.xpath("//header:token", "header" => "http://www.fedora.info/definitions/1/0/types/").children
+    if header.length == 1
+      return header[0].text
```
awead: consolidate:
`return ... if ...`

app/models/migrate_audit_fedora3.rb
```diff
@@ -0,0 +1,130 @@
...
+    docs.each do |doc|
+      if doc.to_s.start_with?(@namespace + ":")
+        yield doc.text
+      end
+    end
+  end
+
+  def extract_session_token(xml)
+    # Get the session token. This is what allows us to fetch the next batch 
+    # of PIDs in the next execution of the query.
```
awead: Ruby style usually puts the comments and explanation above the method call.

app/models/migrate_audit_fedora3.rb
```diff
@@ -0,0 +1,130 @@
...
+    header = xml.xpath("//header:token", "header" => "http://www.fedora.info/definitions/1/0/types/").children
+    if header.length == 1
+      return header[0].text
+    end
+  end
+
+  def extract_title(xml)
+    # Extract the title from the DC data stream
```
awead: comments about the method go above the method call.

app/models/migrate_audit_fedora3.rb
```diff
@@ -0,0 +1,130 @@
...
+    xml_set = xml.xpath("//dc:title", "dc" => "http://purl.org/dc/elements/1.1/")
+    unless xml_set.empty?
```
awead: `xml.xpath("//dc:title", "dc" => "http://purl.org/dc/elements/1.1/").each` makes the unless block unnecessary.

app/models/migrate_audit_fedora3.rb
```diff
@@ -0,0 +1,130 @@
...
+      if xml_set.first.children.count > 0
+        return xml_set.first.children[0].text
```
awead: `return ... if ...`

app/models/migrate_audit_fedora3.rb
```diff
@@ -0,0 +1,130 @@
...
+      end
+    end    
+  end
+
+  def extract_model(xml)
+    # Extract the "has_model" (GenericFile, Batch, Collection) from the RELS-EXT data stream
+    xml_set = xml.xpath("//ns1:hasModel", "ns1" => "info:fedora/fedora-system:def/model#" )
+    unless xml_set.empty?
+      xml_node = xml_set.first
```
awead: `xml.xpath("//ns1:hasModel", "ns1" => "info:fedora/fedora-system:def/model#" ).first` would be nil or the thing you want.

app/models/migrate_audit_fedora3.rb
```diff
@@ -0,0 +1,130 @@
...
+      if xml_node.attributes.key?("resource")
+        return xml_node.attributes["resource"].value
+      end
```
awead:
```
xml_node = xml.xpath("//ns1:hasModel", "ns1" => "info:fedora/fedora-system:def/model#" ).first
unless xml_node.nil?
  return xml_node.attributes["resource"].value if xml_node.attributes.key?("resource")
end
```

but the if statement may be unnecessary and NG would just return nil, so

```
return xml_node.attributes["resource"].value unless xml_node.nil?
```

app/models/migrate_audit_fedora4.rb
```diff
@@ -0,0 +1,97 @@
...
+    f4_model = model_from_resource f4_obj
+    f3_model = normalized_f3_model(f3_obj.f3_model)
+    result = AuditResult.new(f3_obj.id, f3_obj.f3_pid, f4_url, "OK")
+    if f4_obj == nil
+      result.status = "Not found" 
+    elsif f3_model != f4_model 
+      result.status = "Models mismatch. Expected: #{f3_model} but found: #{f4_model}" 
+    end
+    result
+  end
```
awead, line 29: There's a lot to tease out here. You could opt to create a new Auditor class to handle the sticky logic. Otherwise, here's what I came up with:

``` ruby
def audit(f3_objects)
  f3_objects.each do |f3_obj|
    @result = AuditResult.new(f3_obj.id, f3_obj.f3_pid, get_f4_url(f3_obj.f3_pid), nil)
    @result.status = audit_one(f3_obj)
    yield @result
  end
end

private

def audit_one(f3_obj)
  f4_obj = fedora_get_metadata(result.f4_id)
  return "Not found" if f4_obj.nil? 
  f4_model = model_from_resource f4_obj
  f3_model = normalized_f3_model(f3_obj.f3_model)
  return "Models mismatch. Expected: #{f3_model} but found: #{f4_model}" if f3_model != f4_model 
  "OK"
end

def get_f4_url pid
  [@fedora_url, pid_to_uri(pid)].join("/")
  # or consider using URL constructor methods
end
```

app/models/migrate_audit_fedora4.rb
```diff
@@ -0,0 +1,97 @@
...
+  def normalized_f3_model(f3_model)
+    # f3 models are stored like this: "info:fedora/afmodel:Batch"
+    # and we only care about the last part ("Batch" in this case)
+    f3_model.split(':').last
+  end
+
+  def pid_to_uri(pid)
+    id = pid.split(':').last
+    raise "Cannot detect ID from PID #{pid}" unless id.length >= 9
+    id[0..1] + "/" + id[2..3] + "/" + id[4..5] + "/" + id[6..7] + "/" + id
```
awead, line 40: Try using .join: `[ id[0..1], id[2..3], id[4..5],id[6..7], id ].join("/")`

app/models/migrate_audit_fedora4.rb
```diff
@@ -0,0 +1,97 @@
...
+  end
+
+  def model_from_resource(triples)
+    return nil if triples.nil?
```
awead: The above correction should make this line unnecessary because we will have returned "Not Found" before we call this method.

app/models/migrate_audit_fedora4.rb
```diff
@@ -0,0 +1,97 @@
...
+  #     <URI> <info:fedora/fedora-system:def/model#hasModel> "GenericFile"^^<http://www.w3.org/2001/XMLSchema#string> .
+  #
+  # Children come in the form:
+  #     <URI> <http://www.w3.org/ns/ldp#contains> <CHILD-URI-1> .
+  #     <URI> <http://www.w3.org/ns/ldp#contains> <CHILD-URI-2> .
+  #
+  def parse_triples response
+    model = nil
+    children = []
+    start = Time.now
```
awead: did you put this in and forget to take it out? I can't see where you're calling it again.

app/models/migrate_audit_fedora4.rb
```diff
@@ -0,0 +1,97 @@
...
+          model = has_model[0,caret].gsub('"', '')
+        end
+      elsif predicate == "<http://www.w3.org/ns/ldp#contains>"
+        # get the list of children of the object
+        child_url = tokens[2].gsub("<", "").gsub(">", "")
+        children.push child_url
+      end
+    end
+    {model: model, children: children}
+  end
```
awead, line 79: This is screaming refactor to me, but it's too late in the day to grok :smile:

app/models/migrate_audit_fedora4.rb
```diff
@@ -0,0 +1,97 @@
...
+
+  def fedora_get_metadata(url)
+    headers = { "Accept" => "application/n-triples" }
+    uri = URI.parse(url  + "/fcr:metadata")
+    response = Net::HTTP.start(uri.hostname, uri.port) { |http|
```
awead: this is the second time you're called Net::HTTP in this commit. You could roll it out into a module but that's probably overkill. :stars:

lib/tasks/premigrate.rake
```diff
@@ -98,5 +103,47 @@ namespace :pre_migrate do
...
     end
   end
+
+  # This task is mostly for testing purposes to make sure we get all PIDs from Fedora 3.
+  # To allow this task to be run from a version of the code that targets Fedora 3 
+  # or Fedora 4 we must pass the file where the Fedora 3 settings are stored.
+  desc "Fetches PIDs for all objects in Fedora 3 for our namespace"
+  task "get_pids", [:fedora3_yml] => :environment do |cmd, args|
+    file_name = args[:fedora3_yml]
+    abort("Must specify a Fedora 3 YAML file") if file_name.nil?
```
awead, line 113: raise insead? Never seen abort... but it could be just me.

spec/models/migrate_audit_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    results = []
+    audit.audit(f3_data) do |result|
+      results.push result
+    end
+    results
+  end
+
+  before do 
+    f4_file_ok = GenericFile.create(id: "111xyzfile") { |file| file.apply_depositor_metadata('dmc') }
+    f4_file_bad = GenericFile.create(id: "333xyzfile") { |file| file.apply_depositor_metadata('dmc') }
```
awead: I don't think you need the variables if you've not doing anything with them later. Otherwise, use `let()`

spec/models/migrate_audit_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  it 'should detect a file migrated correctly' do
+    migrated = results.select {|r| r.f3_pid == f3_file_migrated.f3_pid}.first
+    expect(migrated.status).to eq("OK")
+    expect(migrated.f4_id).to eq("#{fedora_url}/11/1x/yz/fi/111xyzfile")
+  end
```
awead, line 35: The rspec-way of doing this is to wrap each of these it blocks in a context:

```
context "with correctly migrated files" do
  subject { results.select {|r| r.f3_pid == f3_file_migrated.f3_pid}.first }
  it "should be 'OK' do
    expect(subject.status).to eq("OK")
  end
  it "should have a valid fedora 4 url" do
    expect(subject.f4_id).to eq("#{fedora_url}/11/1x/yz/fi/111xyzfile")
  end
end
```

but this is way :eyeglasses:

awead:
@cam156 @mjgiarlo so I went :banana: :banana: :banana: with the comments. I want to say that overall, I'm :100: :+1: with this _as is_. I was just exercising a bit here, in case @hectorcorrea wanted some deeper level comments. Feel free to provide additional suggestions.

The only thing I'm not sure of are all the changes to `schema.rb` but that's just me. If we want to pull this code out eventually, that might make the schema a little weird to deal with later, but I'm not too familiar with this on the Rails side.

It's now :hourglass: and time for :cocktail:

app/models/migrate_audit_fedora3.rb
mjgiarlo, line 98: Yup, that's exactly the convention we've been following, which is more or less this: https://github.com/bbatsov/ruby-style-guide

app/models/migrate_audit_fedora4.rb
mjgiarlo, line 29: Just one thing to add: methods that begin with "get" or "set" are generally avoided in Ruby as a matter of convention.  Again, just a general guidelines.  If you want to get an F4 URL back, you call the method `f4_url`.

lib/tasks/premigrate.rake
mjgiarlo, line 113: I've never seen abort either.  What's that do? :)

spec/models/migrate_audit_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  let(:f3_file_migrated) { FactoryGirl.create(:f3_file_migrated) }
+  let(:f3_file_not_migrated) { FactoryGirl.create(:f3_file_not_migrated) }
+  let(:f3_file_migrated_wrong) { FactoryGirl.create(:f3_file_migrated_wrong) }
+  let(:f3_data) { [f3_file_migrated, f3_file_not_migrated, f3_file_migrated_wrong] }
+
+  let(:audit) { MigrateAuditFedora4.new(fedora_url, credentials[:user], credentials[:password]) }
+
+  let(:results) do 
+    results = []
```
mjgiarlo, line 15: Not necessary, but here's a little pattern I use sometimes to avoid code blocks where you set an array/hash variable, loop over something and pop stuff into the array/hash, and return the variable on a line by itself:

``` ruby
[].tap do |results|
  audit.audit(f3_data) do |result|
    results.push result
  end
end
```

You could argue that using `#tap` is a little less clear to less experienced Rubyists, and I think that's valid. I dig this pattern because it means I'm typing "results" fewer times. :)

spec/models/migrate_audit_spec.rb
mjgiarlo: :+1: to `let`. Set variables via `let` and use `before` for triggering/setting up behavior.

spec/models/migrate_audit_spec.rb
mjgiarlo, line 35: This is a guideline we follow inconsistently, but I prefer eschewing the word "should" in test descriptions because it's implicit.  This comes from the RSpec style guide here: http://betterspecs.org/#should  Instead I'd reword as:

``` ruby
it 'returns OK' ...
it 'has a valid Fedora 4 URL' ...
```

mjgiarlo:
:+1: to @awead's comments! These are useful pointers. Thanks for whipping this up and working through the PR process, @hectorcorrea

lib/tasks/premigrate.rake
hectorcorrea, line 113: Abort terminates the process with a returns value of false (indicating not a successful run) and sends to STDERR the error message.

spec/models/migrate_audit_spec.rb
hectorcorrea: Good catch.

spec/models/migrate_audit_spec.rb
hectorcorrea, line 35: Good to know. I've updated the code.

spec/models/migrate_audit_spec.rb
hectorcorrea, line 15: I'll stay with the newbie approach. I am not too comfortable with (obscure) Ruby tricks just yet :)

app/models/migrate_audit.rb
hectorcorrea: Cool. I like your approach better. Done.

app/models/migrate_audit.rb
hectorcorrea: Done.

app/models/migrate_audit.rb
hectorcorrea: Done.

app/models/migrate_audit.rb
hectorcorrea: Indeed. Done.

app/models/migrate_audit_fedora3.rb
hectorcorrea: They are now gone!

app/models/migrate_audit_fedora3.rb
hectorcorrea: Moved the comment to the top.

app/models/migrate_audit_fedora3.rb
hectorcorrea, line 39: I missed this. Future enhancement if somebody feels strongly about it :)

app/models/migrate_audit_fedora3.rb
hectorcorrea: Done.

app/models/migrate_audit_fedora3.rb
hectorcorrea: Done.

app/models/migrate_audit_fedora3.rb
hectorcorrea, line 98: Good catch. Fixed.

hectorcorrea:
@awead @mjgiarlo I've implemented most of the suggestions that @awead provided. Thanks a lot for the comprehensive feedback. I also removed a few hash rockets that you guys missed ;) 

The suggestions that I didn't implement are the ones that were along the lines of "reduce these the 3 lines to a single line" when I thought the assignment of the variables line by line made the code clearer (and easier to debug) than the suggested single line approach. 

I also left a few string concatenations in place (instead of replacing them with string interpolation) when I thought the concatenation made the operation easier to read.

mjgiarlo:
:clap:

PR #80, opened 2015-02-28
Collection size is not zero bytes, ref #9753

awead:
This was fixed in Sufia, but I avoided putting the follow-up test as a feature, since that would require uploading files and would slow down the test suite. Instead, the size property on the file is mocked in a unit environment, similarly to the way it is in the Sufia test.

### DIFF
--- a/spec/models/collection_spec.rb
+++ b/spec/models/collection_spec.rb
@@ -1,8 +1,31 @@
 require 'spec_helper'
 
 describe Collection do
+  
   describe "#bytes" do
-    subject { described_class.new.bytes }
-    it { is_expected.to eq 0 }
+    context "with no attached files" do
+      subject { described_class.new.bytes }
+      it { is_expected.to eq 0 }
+    end
+
+    context "with attached files" do
+      let(:file) do
+        GenericFile.create(title: ['Some title']) { |f| f.apply_depositor_metadata("agw") }
+      end
+      let(:file2) do
+        GenericFile.create(title: ['Some other title']) { |f| f.apply_depositor_metadata("agw") }
+      end
+      let(:collection) do
+        Collection.create(title: "My collection") { |c| c.apply_depositor_metadata("agw") }
+      end
+      let(:mock_file) { double("content", size: "100")}
+      before do
+        collection.members = [file, file2]
+        allow_any_instance_of(GenericFile).to receive(:content).and_return(mock_file)
+      end
+      subject { collection.bytes }
+      it { is_expected.to eq 200 }
+    end
+  
   end
 end


spec/models/collection_spec.rb
```diff
@@ -1,8 +1,43 @@
...
-    subject { described_class.new.bytes }
-    it { is_expected.to eq 0 }
+    context "with no attached files" do
+      subject { described_class.new.bytes }
+      it { is_expected.to eq 0 }
+    end
+
+    context "with attached files" do
+      let(:file) do
+        GenericFile.new.tap do |f|
```
mjgiarlo: If you use `GenericFile.create do |f|` here, you can eliminate the `f.save!` below.

You can also do `GenericFile.create(title: ['Some title']) do |f|` and eliminate the title setting below.

spec/models/collection_spec.rb
```diff
@@ -1,8 +1,43 @@
...
+          f.title = ["Some title"]
+          f.apply_depositor_metadata("agw")
+          f.save!
+        end
+      end
+      let(:file2) do
+        GenericFile.new.tap do |f|
```
mjgiarlo: Ditto.

spec/models/collection_spec.rb
```diff
@@ -1,8 +1,43 @@
...
+          f.title = ["Some other title"]
+          f.apply_depositor_metadata("agw")
+          f.save!
+        end
+      end
+      let(:collection) do
+        Collection.new.tap do |c|
```
mjgiarlo: Ditto.

spec/models/collection_spec.rb
```diff
@@ -1,8 +1,43 @@
...
+          c.title = "My collection"
+          c.apply_depositor_metadata("agw")
+          c.save!
+        end
+      end
+      let(:mock_file) { double("content", size: "100")}
+      subject do
```
mjgiarlo: Though the current code works, I think it's cleaner to have one-line `subject`s with setup happening in a `before` block. Like:

``` ruby
before do
  allow_any_instance_of(GenericFile).to receive(:content) { mock_file }
  collection.members = [file1, file2]
end

subject { collection.bytes }
```

spec/models/collection_spec.rb
carolyncole: @mjgiarlo  You will need a create! to replicate the the save! correct?   Also this seems like a nice to have not required for the PR.

spec/models/collection_spec.rb
carolyncole: @mjgiarlo I agree that this is a bit confusing about what the subject is going to get set to.  I do think it is a nice to have though...

spec/models/collection_spec.rb
mjgiarlo: Yes, indeed, a nice-to-have. As currently written this does not introduce much if any technical debt since it's just specs and in re lines of code.

spec/models/collection_spec.rb
mjgiarlo: I think we've seen enough of this pattern to figure it out, rue. I leave it to @awead if he wants to make this change or leave it.

awead:
@cam156 @mjgiarlo updated!

PR #81, opened 2015-03-05
Make sure we don't create a batch with the same ID 

hectorcorrea:
...as an another existing object (e.g. we don't overwrite a GenericFile)

### DIFF
--- a/lib/tasks/premigrate.rake
+++ b/lib/tasks/premigrate.rake
@@ -61,7 +61,13 @@ namespace :pre_migrate do
     missing_batches = []
     batches_in_files.each do |id|
       unless all_batches.include? id
-        missing_batches.push(id)
+        if ActiveFedora::Base.exists?(pid: id)
+          # An object with this PID already exists 
+          # (it might exist as something other than a Batch)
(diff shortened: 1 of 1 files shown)


lib/tasks/premigrate.rake
```diff
@@ -61,7 +61,13 @@ namespace :pre_migrate do
     missing_batches = []
     batches_in_files.each do |id|
       unless all_batches.include? id
-        missing_batches.push(id)
+        if ActiveFedora::Base.find(pid: id).count == 0
```
mjgiarlo: How about:

``` ruby
  if ActiveFedora::Base.exists?(pid)
    # This is an indication that an object with this PID already
    # exists (it might exists as something other than a Batch
    logger.error "Skipping #{id} because it already exists!"
  else
    missing_batches.push(id)
  end
```

:grey_question:  So you don't have to run a `find` and a `count`.

hectorcorrea: @mjgiarlo good idea. I've updated it.

mjgiarlo:
Code looks good but Travis is :unamused:

PR #86, opened 2015-03-20
Refactoring query_solr to use ActiveFedora solr.

carolyncole:


lib/tasks/scholarsphere.rake
```diff
@@ -71,25 +69,18 @@ namespace :scholarsphere do
...
-  end
-
   desc "Characterize all files"
   task characterize: :environment do
     # grab the first increment of document ids from solr
-    resp = query_solr(q:"{!lucene q.op=AND df=text}id:#{ScholarSphere::Application.config.redis_namespace}\\:* has_model_s:*GenericFile*" )
+
+   resp = ActiveFedora::SolrService.instance.conn.get "select",
+              params:{ fl:['id'], fq: "#{ Solrizer.solr_name("has_model", :symbol)}:GenericFile"}
+    puts resp
```
awead, line 83: I was hoping to rework these bits, and all the other instances, to use https://github.com/projecthydra/sufia/blob/master/app/search_builders/sufia/search_builder.rb#L20
as well as other methods from this class, but we could ticket this for later.

PR #92, opened 2015-03-31
Adding rack-maintenance to allow for a maintenance page to be served eas...

carolyncole:
...ily before the app gets hit

I'm wondering if we could use this to put up a maintenance page while we are upgrading the server.

The only down side I see is that we would not be able to testing without removing the maintenance page, but we would have to turn on the web apps test anyway

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,6 +29,7 @@ gem 'therubyracer'
 gem 'uglifier'
 gem 'whenever'
 gem 'yaml_db'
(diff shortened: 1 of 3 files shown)


awead:
Are we still doing this or just closing it off via ip?

carolyncole:
I think we may still need to do this.  Unfortunately the load balancer is changing the in coming IP

PR #113, opened 2015-04-03
Adding my_collections_controller to override sufia and show only collect...

### LINKED ISSUE #93: Collections show all collections
The collection list on my dashboard shows the entire lists of system collections not just my collections.

carolyncole:
...ions from the current user fixes #93

### DIFF
--- a/spec/features/dashboard/dashboard_collections_spec.rb
+++ b/spec/features/dashboard/dashboard_collections_spec.rb
@@ -6,6 +6,15 @@ describe 'Dashboard Collections:', :type => :feature do
 
   let!(:current_user) { create :user }
 
+  let(:jill) { create :jill }
+  let!(:collection) do
+    Collection.new.tap do|col|
+      col.apply_depositor_metadata(jill.user_key)
+      col.title = "jill collection"
+      col.save!
+    end
+  end
+
   before do
     sign_in_as current_user
     go_to_dashboard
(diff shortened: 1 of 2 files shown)


spec/features/dashboard/dashboard_collections_spec.rb
```diff
@@ -6,6 +6,15 @@
 
   let!(:current_user) { create :user }
 
+  let(:jill) { create :jill }
+  let!(:collection) do
+    Collection.new.tap do|col|
```
mjgiarlo, line 11: You can tighten this up to:

``` ruby
Collection.create(title: 'jill collection') do |col|
  col.apply_depositor_metadata(jill.user_key)
end
```

spec/features/dashboard/dashboard_collections_spec.rb
```diff
@@ -25,6 +34,7 @@
 
   specify 'collections are displayed in the Collections list' do
     expect(page).to have_content "My collection"
+    expect(page).not_to have_content "jill collection"
```
mjgiarlo, line 37: Since the intent is different here, could you break this into its own test?

``` ruby
  specify 'only collections deposited by the logged-in user are displayed' do
    expect(page).not_to have_content "jill collection"
  end
```

carolyncole, line 37: That would then make the entire feature test slow by spinning up the app again.  That is why the feature test are sooooo slow.  My intent if we were not in a code freeze would be to combine all the test into one.  So the answer is no I definitely can not.

PR #114, opened 2015-04-03
Overwrite Sufia's version of this controller...

### LINKED ISSUE #104: Batch Edit stuck when changing permissions
- Go to dashboard
- Select 2 files
- Click Edit Selected
- Click Permissions
- Change the permissions (e.g. from public to private)
- Click Save Changes
- The system gets stuck and the permissions are _not updated_

screen shot 2015-04-02 at 4 14 08 pm

Test case https://scm.dlt.psu.edu/issues/3726

hectorcorrea:
...to handle the case where no generic file data is received (see check on generic_file_parameters method.) This will need to be fixed in Sufia as well in order for us to remove this entire file from ScholarSphere.

Fixes #104

### DIFF
--- a/app/controllers/concerns/sufia/batch_edits_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/batch_edits_controller_behavior.rb
@@ -1,0 +1,100 @@
+module Sufia
+  module BatchEditsControllerBehavior
+    extend ActiveSupport::Concern
+    include Sufia::Breadcrumbs
+
+    included do
+      layout "sufia-one-column"
+      before_filter :build_breadcrumbs, only: :edit
+    end
+
+    def edit
+       super
+       @generic_file = ::GenericFile.new
+       @generic_file.depositor = current_user.user_key
+       @terms = terms - [:title, :format, :resource_type]
+
+       h  = {}
+       @names = []
+       permissions = []
+
+       # For each of the files in the batch, set the attributes to be the concatination of all the attributes
+       batch.each do |doc_id|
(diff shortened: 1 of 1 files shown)


mjgiarlo:
Should we override the controller behavior or add an app/controllers/batch_edits_controller.rb to ScholarSphere, include the controller behavior, and override whatever methods need overriding? The Alberta folks are doing this-- https://github.com/ualbertalib/HydraNorth/blob/master/app/controllers/batch_edits_controller.rb -- and we do the same for the collections controller -- https://github.com/psu-stewardship/scholarsphere/blob/develop/app/controllers/collections_controller.rb.

(It doesn't matter too much if we plan to push this into Sufia soon after our sphere release, I suppose, but I thought I'd ask!)

carolyncole:
@mjgiarlo Yes we should do this locally because we are in a code freeze.  We upgraded to the latest sufia last week and now we are in a bad state because of the new bugs introduced by the upgrade.  Upgrading to another version of sufia would leave us open to more bugs.  We do not have time in the schedule to do that again.

hectorcorrea:
@mjgiarlo I've submitted a separate PR to Sufia for the "correct" fix. See: https://github.com/projecthydra/sufia/pull/1027 As you see in the Sufia PR the fix is just to initialize one of the parameters rather than passing a null value.

mjgiarlo:
@cam156 I didn't suggest that we not do this locally. I asked if we should be overriding the behavior or the controller itself. But if what has already been merged works then, as I wrote in parentheses, it doesn't matter too much since this is a short-term change.

PR #115, opened 2015-04-06
Allow only editable collections in the collection list. fixes #109

### LINKED ISSUE #109: When I add files to a collection I see all the collections
When I go to my files
And I check files 
And I click add files to collection
Then I should only see my collections in the list

1 - add to collection view all collections

This is related to #93

carolyncole:
Overriding to find only collections with edit access.  Sufia currently finds any collections (read, or edit access).

Doing this here for the moment to facilitate the release.  Need to have a community discussion to see if this is need/wanted in sufia.  I can not see why a a user might select a collection they do not have edit access to, but there may be a change in sufia I am unaware of.

### DIFF
--- a/app/controllers/my_controller.rb
+++ b/app/controllers/my_controller.rb
@@ -1,0 +1,6 @@
+class MyController < ApplicationController
+  include Sufia::MyControllerBehavior
+  skip_before_filter :find_collections, only: :index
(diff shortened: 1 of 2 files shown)


mjgiarlo:
:+1:

PR #117, opened 2015-04-07
Setting edit access for shares so we only see files the user has edit ac...

### LINKED ISSUE #94: Files shared with me includes everyone's files
The list of files shared with me shows all the files on the system not just the files shared with me.

carolyncole:
...cess to.  fixes #94 

Adding overrides to searchbuilder to add edit filter (self.discovery_perms = [:edit]) to show_only_shared_files

Adding shares_controller to change the order of the solr filers so the only_shares comes before the access control ones

### DIFF
--- a/app/controllers/my/shares_controller.rb
+++ b/app/controllers/my/shares_controller.rb
@@ -1,0 +1,22 @@
+module My
+  class SharesController < MyController
+
+    self.search_params_logic = [
+      :show_only_shared_files,
+      :show_only_generic_files
+    ] + self.search_params_logic
+
+
+    def index
+      super
+      @selected_tab = :shared
+    end
+
+    protected
+
+    def search_action_url *args
+      sufia.dashboard_shares_url *args
+    end
+
+  end
+end
--- a/app/search_builders/sufia/search_builder.rb
+++ b/app/search_builders/sufia/search_builder.rb
@@ -1,0 +1,53 @@
+module Sufia::SearchBuilder
+
+  include BlacklightAdvancedSearch::AdvancedSearchBuilder
+  include Hydra::Collections::SearchBehaviors
+  include Hydra::Controller::SearchBuilder
+
+  def show_only_collections(solr_parameters)
+    solr_parameters[:fq] ||= []
+    solr_parameters[:fq] += [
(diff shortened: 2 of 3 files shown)


app/controllers/my/shares_controller.rb
```diff
@@ -0,0 +1,22 @@
+module My
+  class SharesController < MyController
+
+    self.search_params_logic = [
+      :show_only_shared_files,
+      :show_only_generic_files
+    ] + self.search_params_logic
+
```
awead, line 8:
``` ruby
self.search_params_logic << [ :show_only_shared_files, :show_only_generic_files]
```

does that do the same thing?

mjgiarlo, line 8: No, the shovel will not work here because it will add an array to search_params_logic rather than two elements (shovel is `#append`, plus is `#concat`).  I think the common pattern here is:

``` ruby
self.search_params_logic += [:show_only_shared_files, :show_only_generic_files]
```

carolyncole, line 8: Yes and that appends the methods.  I am specifically putting my methods first so I can influence other routines in the list.  specifically add_access_controls_to_solr_params.  That routine defaults to all access read & edit unless you tell it otherwise.  The old way things worked before search builders entered the arena these lines set the level: https://github.com/projecthydra/sufia/blob/master/app/controllers/concerns/sufia/my_controller_behavior.rb#L60-L62

carolyncole, line 8: @awead @mjgiarlo I'm not certain this is a the long term fix.  I think that allowing the search builder to be assigned a access_level would be the best way, but I could not determine how to accomplish that.  This is an interim solution to patch the fact that blacklight was changed and now our code does not work.

mjgiarlo, line 8: :ok:

mjgiarlo, line 8: No disagreement! Deviating from the pattern above makes sense if ordering of rules is desired, and right now given lack of better options and a shortage of time, it is. I totally get it. Mostly I wanted to say that plus is a better fit than shovel in this context.

PR #120, opened 2015-04-08
Update FedoraMigrate

awead:


carolyncole:
We are getting a large set of changes to base gems in with this PR.  Have you looked at all the gems that are being updated to make sure there are no breaking changes?  I wonder if we should pin all those versions to the old versions.

awead:
@cam156 I took the minimal approach and only make the smallest change to get the tests to pass. All of the changes are patch-level, which should only be fixes and security improvements.

I'd also advocate for only pinning the versions that are causing problems, ie. AF, that way we're getting the latest versions of gems when we need them, and we'll know which specific gem is causing the problem when it happens.

carolyncole:
@awead so the test will fail with this branch? https://github.com/psu-stewardship/scholarsphere/compare/migration-testing2?expand=1

That branch only changes the fedora-migrate gem, not other dependent gems.

Not everyone follow semantic versioning.  We might get some surprises...

awead:
@cam156 I'd need to test that by running a migration locally.

awead:
@cam156 updated with your changes from migration-test2 branch, and rebased against develop so it's up to date.

Issue #119, opened 2015-04-08
Creator Field Default


carolyncole:
Creator Field is being defaulted to the depositor
Per Patricia this should be in the format "lastname, firstname"
We are putting in "firstname last" name format
We should fix the defaulting  to the proper format.

carolyncole:
Need to override this line: https://github.com/projecthydra/sufia/blob/master/app/controllers/concerns/sufia/batch_controller_behavior.rb#L38

PR #123, opened 2015-04-08
adding namae parser for names to reverse the order of the creator.  fixe...

carolyncole:
...s #119

### DIFF
--- a/app/controllers/batch_controller.rb
+++ b/app/controllers/batch_controller.rb
@@ -1,0 +1,10 @@
+class BatchController < ApplicationController
+  include Sufia::BatchControllerBehavior
+
+  # override sufia default to put creator in format PSU expects, lastname, first name
+  def edit_form
+    parsed = Namae::Name.parse(current_user.name)
+    generic_file = ::GenericFile.new(creator: [parsed.sort_order], title: @batch.generic_files.map(&:label))
+    edit_form_class.new(generic_file)
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -30,6 +30,7 @@ gem 'uglifier'
 gem 'whenever'
 gem 'yaml_db'
 gem 'rack-maintenance'
+gem 'namae', '0.9.3'
 
 group :development, :test do
   gem 'jettywrapper'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -360,6 +360,7 @@ GEM
     multi_xml (0.5.5)
     multipart-post (2.0.0)
     mysql2 (0.3.18)
+    namae (0.9.3)
     nest (1.1.2)
       redis
     net-http-persistent (2.9.4)
@@ -726,6 +727,7 @@ DEPENDENCIES
   kaminari!
   launchy
   mysql2 (~> 0.3.17)
+  namae (= 0.9.3)
   nest
   newrelic_rpm
   passenger
--- a/spec/controllers/batch_controller_spec.rb
+++ b/spec/controllers/batch_controller_spec.rb
@@ -1,0 +1,20 @@
+require 'spec_helper'
+
+describe BatchController, type: :controller do
+  routes { Sufia::Engine.routes }
+  let (:subject) {BatchController.new}
+  let (:current_user) { User.new(display_name: "ALBERT EDWARD MATYASOVSKY JR") }
+  let (:batch) {Batch.new}
+
+  describe "#edit_form" do
+    before do
+      allow(subject).to receive(:current_user).and_return(current_user)
+      subject.instance_variable_set("@batch", batch)
+    end
+    it "creates an edit form with the creator in last name first ordering" do
+      expect(subject.edit_form[:creator]).to eq(["Matyasovsky, Jr, Albert Edward"])
+    end
+  end
+  
+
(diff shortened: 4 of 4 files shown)


app/controllers/batch_controller.rb
```diff
@@ -0,0 +1,13 @@
+class BatchController < ApplicationController
+  include Sufia::BatchControllerBehavior
+
+  # override sufia default to put creator in format PSU expects, lastname, first name
+  def edit_form
+    parsed = Namae::Name.parse(current_user.name)
+    creator = parsed.family+", "+parsed.given
+    puts parsed.inspect
```
awead: remove this?

app/controllers/batch_controller.rb
```diff
@@ -0,0 +1,13 @@
+class BatchController < ApplicationController
+  include Sufia::BatchControllerBehavior
+
+  # override sufia default to put creator in format PSU expects, lastname, first name
+  def edit_form
+    parsed = Namae::Name.parse(current_user.name)
+    creator = parsed.family+", "+parsed.given
```
awead: I'm assuming the name is coming from LDAP, and there's no way that they would not have a first and a last name.

carolyncole: Unfortunately we do not have first and last name spearately in the system currently.
Into the future we may want to add first and last name if we can get it from the ldap.

awead: Sorry, I meant `family` and `given`. I was wondering if there would ever be any case where those returned `nil` or threw a `NoMethod` error.

mjgiarlo: If either can return `nil`, we should probably use string interpolation here rather than concatenation. Here's why:

```
[1] pry(main)> first_name = "Mike"
=> "Mike"
[2] pry(main)> last_name = nil
=> nil
[3] pry(main)> last_name + ", " + first_name
NoMethodError: undefined method `+' for nil:NilClass
from (pry):3:in `__pry__'
[4] pry(main)> "#{last_name}, #{first_name}"
=> ", Mike"
```

Of course then there's a broader problem of displaying malformed names, but maybe we're OK?  Functionality that inverts names is deceptively simple, but names are actually surprisingly complex. See: http://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/

That said, I think we can live with this implementation for now, and malformed names are preferable to unrescued NoMethodErrors...

awead: Agreed, @mjgiarlo. Also, this information is originating from LDAP, I think, so you can't have just "Mike" for your name. You have to have a family and given name.

mjgiarlo: I agree that that is usually true, and perhaps that's good enough, but that's not always a safe bet. See above URL for more! (And that's assuming too that our LDAP data is flawless...)

To be clear: I'm not saying we need to cover every edge case here, but we ought to be careful (re: the above URL) with our assumptions about names, especially when a single nil can throw an error in a user's face in production.

app/controllers/batch_controller.rb
```diff
@@ -0,0 +1,12 @@
+class BatchController < ApplicationController
+  include Sufia::BatchControllerBehavior
+
+  # override sufia default to put creator in format PSU expects, lastname, first name
+  def edit_form
+    parsed = Namae::Name.parse(current_user.name)
+    creator = "#{parsed.family}, #{parsed.given}"
```
carolyncole: @mjgiarlo @awead better?   I am leaning on a gem to do the parsing.  Yes it would be better to get the fields separately from ldap, but with all the ldap issues we currently have and the timing on the project changing our user model to accommodate other name fields seems more risky than using a gem.

The other option is to not fill it in at all.  We could just remove the defaulting.

mjgiarlo: Even better: let Namae handle the formatting for you (since it already handles lots of cases where only one name or the other exists). See an example here:

``` ruby
[1] pry(main)> name = Namae::Name.parse('Mr. Michael J. Giarlo')
=> #<struct Namae::Name
  family="Giarlo",
  given="Michael J.",
  particle=nil,
  dropping_particle=nil,
  nick=nil,
  appellation="Mr.",
  title=nil>
[2] pry(main)> name.sort_order
=> "Giarlo, Michael J."
[3] pry(main)> Namae::Name.parse('Mr. Giarlo').sort_order                                                                                        
=> "Giarlo"
[4] pry(main)> Namae::Name.parse('Michael').sort_order
=> "Michael"
```

Here's the method you can use: https://github.com/berkmancenter/namae/blob/master/lib/namae/name.rb#L9

carolyncole: @mjgiarlo That does not include the suffix, only the name.  Are we ok with that?

mjgiarlo: @cam156 Doesn't it?

``` ruby
    def sort_order(delimiter = ', ')
      [family_part, suffix, given_part].compact.reject(&:empty?).join(delimiter)
    end
```

I see suffix there!

carolyncole: @mjgiarlo I looked last night, but I missed it because it was in the middle.

PR #124, opened 2015-04-08
Adding a message that show that we are migrating. fixes #121

### LINKED ISSUE #121: Add Screen Block
This issue should be completed on Friday

Add a screen block in so that users will not want to interact with the system.

Make a large red block similar to the current yellow block that says migrating....

carolyncole:
If you are an admin user the warning will not show for you, otherwise you need to scroll down a lot to get to the main page.

screen shot 2015-04-08 at 4 32 46 pm

### DIFF
--- a/app/views/_masthead.html.erb
+++ b/app/views/_masthead.html.erb
@@ -1,4 +1,9 @@
 <div id="masthead" class="container-fluid">
+  <% unless can? :update, FeaturedWork %>
+  <div class="row error">
+    <h2>ScholarSphere is being upgraded and will be back on by Monday April 13th at 7am. Please come back then!</h2>
+  </div>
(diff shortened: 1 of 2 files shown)


app/views/_masthead.html.erb
```diff
@@ -1,4 +1,9 @@
 <div id="masthead" class="container-fluid">
+  <% unless can? :update, FeaturedWork %>
+  <div class="row error">
+    <h2>ScholarSphere is being maintained come back later!!</h2>
```
mjgiarlo: How about:

```
ScholarSphere is being upgraded and will be back on DATE/TIME. Please come back then!
```

carolyncole:
@mjgiarlo updated the text

Issue #90, opened 2015-03-31
Collection edit drop down menu is split


carolyncole:
The far right menu the arrow is split from the menu bar

Looks like there is plenty of room there for the menu to be on the same line...

![screen shot 2015-03-31 at 9 12 51 am](https://cloud.githubusercontent.com/assets/1599081/6919642/db578bd4-d786-11e4-88d7-2361cf9351fd.png)

awead:
@cam156 where is this? What's the url?

awead:
@cam156 ah, nm, found it!

carolyncole:
@awead I saw you fixed this in sufia.  Since we will not be getting a new version of sufia for the fedora 4 release I changed the milestone to the next version.

mtribone:
![screen shot 2015-06-25 at 10 55 12 am](https://cloud.githubusercontent.com/assets/4163828/8357380/aa00a7a6-1b28-11e5-9420-8520d91f7223.png)

PR #128, opened 2015-04-09
Fixes alignment of dropdown in #90

awead:


app/views/collections/_action_menu.html.erb
```diff
@@ -0,0 +1,26 @@
+<% # TODO: this file can be deleted once https://github.com/projecthydra/sufia/pull/1031 is merged %>
```
mjgiarlo, line 1: I wonder if `<%#` might be preferable to `<% #` since the former handles the comment at the ERB layer rather than telling ERB that there is Ruby code to execute (and then it executes nothing)? Minor quibble, no need to hold up the PR.

awead, line 1: @mjgiarlo huh. had no idea about that one. Using `<%#` does keep it out of the rendered html, right?

mjgiarlo, line 1: yep.

PR #130, opened 2015-04-10
Changing contributions to highlighted contributions, so that people do n...

### LINKED ISSUE #100: Contributions tab on profile is confusing
The Contributions tag on the profile should say Highlighted Contributions (or Files)
The tab itself should not include another label

Jim expected Contributions to be a list of all contributions...

screen shot 2015-04-02 at 11 45 22 am

carolyncole:
...ot think it is all the contributions. fixes #100 

Before:
screen shot 2015-04-10 at 7 37 15 am

After you logged in:
screen shot 2015-04-10 at 7 38 09 am

After another user:
screen shot 2015-04-10 at 7 42 18 am

### DIFF
--- a/app/views/users/_contributions.html.erb
+++ b/app/views/users/_contributions.html.erb
@@ -1,0 +1,31 @@
+<div class="tab-pane active" id="contributions">
+  <% if @trophies.count > 0 %>
+    <table>
+      <% @trophies.each do |t| %>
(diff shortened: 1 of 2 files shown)


carolyncole:
At some point in the future this should be made an I18n configuration item in Sufia, so that we do not need to override the entire partial to change the tab labels.

Issue #140, opened 2015-04-13
URLs with "scholarsphere:"


hectorcorrea:
Our logs shows several requests for URLs that include "scholarsphere:" which I understand we supported at one point (for example: /collections/scholarsphere:123) 

Currently scholarsphere throws a "System Error" when it receives these requests. 

We should update ScholarSphere to detect these cases in files, collections, and downloads (are there any others?) and redirect the user to the proper URL without "scholarsphere:" on it. For example HTTP GET /collections/scholarsphere:123 should return an HTTP status 301 (moved permanently) to /collections/123

Here is an example of a page referencing the old URLs: https://www.libraries.psu.edu/psul/socialsciences/docdigiproj/environmentalsci.html

carolyncole:
We should use something similar to what we used to do: https://github.com/projecthydra/sufia/blob/5.0.0-stable/app/controllers/concerns/sufia/files_controller_behavior.rb#L41 and
https://github.com/projecthydra/sufia/blob/5.0.0-stable/app/controllers/concerns/sufia/controller.rb#L15-L17

Except we should remove the namespace.  

I marked this as high priority since we know of users that need this.

carolyncole:
This fix has been put into production.

PR #141, opened 2015-04-14
Redirect URLs with legacy scholarsphere prefix in the URL

hectorcorrea:
to the new URL format without it.

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -1,6 +1,12 @@
 class CollectionsController < ApplicationController
   include Sufia::CollectionsControllerBehavior
 
+  prepend_before_filter only: [:show, :edit] do  
+    handle_legacy_url_prefix do |new_id|
+      redirect_to collections.collection_path(new_id), status: :moved_permanently
+    end
+  end 
+
   def presenter_class
     ::CollectionPresenter
   end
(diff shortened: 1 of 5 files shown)


hectorcorrea:
Fixes #140

app/controllers/collections_controller.rb
```diff
@@ -1,6 +1,12 @@
 class CollectionsController < ApplicationController
   include Sufia::CollectionsControllerBehavior
 
+  prepend_before_filter only: [:show, :edit] do  
+    handle_legacy_url_prefix do |new_id|
+      redirect_to collections.collection_path(new_id), status: :moved_permanently
```
carolyncole: Since this is a one line block you could use the single line syntax
handle_legacy_url_prefix {|new_id| redirect_to collections.collection_path(new_id), status: :moved_permanently}
(not a necissary change)

carolyncole:
:+1: Looks good to me.  It would be nice to have a test, to be certain we do not lose this functionality with updates.

app/controllers/collections_controller.rb
hectorcorrea: Done.

hectorcorrea:
@cam156 I added a test for the collections controller. I didn't add a test for files and downloads since they would be almost identical, but I could if you think that is necessary.

Issue #146, opened 2015-04-20
Generic File show page should use solr instead on Fedora


carolyncole:
The GenericFile show page should use solr for showing the document instead of using the fedora document.
Here is where we used to load from solr before F4
https://github.com/psu-stewardship/scholarsphere/blob/develop/app/controllers/generic_files_controller.rb#L22-L29

Here is the removed code:
https://github.com/psu-stewardship/scholarsphere/blob/v2.1.1/app/controllers/generic_files_controller.rb#L12-L15

Here is some code in sufia that may help: https://github.com/projecthydra/sufia/issues/711

awead:
Substituting a solr-based GenericFile object as opposed to an ActiveFedora one raises a problem here:
https://github.com/projecthydra/sufia/blob/master/sufia-models/app/services/sufia/generic_file_audit_service.rb#L33
because the audit service needs access to the file's version information.

I believe the best way to fix this is to add version information to ActiveFedora::LoadableFromJson::SolrBackedMetadataFile so that Sufia's audit service can remain unchanged and you can get the same results from the audit service whether you use a full AF object or just one loaded from Solr.

awead:
The other problem is characterization information. We'd need to override this view:
https://github.com/projecthydra/sufia/blob/master/app/views/generic_files/_show_details.html.erb#L20
and replace the call to .characterization_terms with something else.

PR #149, opened 2015-04-22
Potential solution to solr-only show view

awead:
Address a solr-only view in #146. This still needs some work, namely that the AuditService ought to have more separate concerns when it comes to getting the status of an audit versus performing one.

This work should eventually make it back into Sufia as well.


awead:
This is good enough for now. I'm curious to see what performance improvements we see when this is deployed to production. If it's a good boost, I'll rework this into Sufia.

carolyncole:
This should really be a PR to release-2.2.4 not develop.  (in theory develop could be getting new features added)

Issue #125, opened 2015-04-09
scholarSphere dashboard controller is duplicate code


carolyncole:
I believe that ScholarSphere dashboard_controller.rb can be removed.
Do not want to change anything I do not need to now after code freeze.
The file should be removed and the system retested.

These lines: https://github.com/psu-stewardship/scholarsphere/blob/develop/app/controllers/dashboard_controller.rb#L5-L6
Noe live in sufia: https://github.com/projecthydra/sufia/blob/master/app/controllers/concerns/sufia/dashboard_controller_behavior.rb#L37-L38

PR #150, opened 2015-04-22
Remove DashboardController, addresses #125

awead:
DashboardController is now fully in Sufia. So remove it and its unit tests, but leave a feature test for future-proofing.


carolyncole:
This did not require an upgrade to sufia in the Gemfile?  Thanks for following up with this.

awead:
Nope. The DashboardController hasn't changed much.

PR #151, opened 2015-04-22
Load GenericFile from Solr for show view #146

### LINKED ISSUE #146: Generic File show page should use solr instead on Fedora

awead:
Fixes #146

### DIFF
--- a/app/views/generic_files/_show_details.html.erb
+++ b/app/views/generic_files/_show_details.html.erb
@@ -1,0 +1,21 @@
+<h2>File Details</h2>
+<dl class="dl-horizontal file-show-term file-show-details">
+  <dt>Depositor</dt>
+  <dd itemprop="accountablePerson" itemscope itemtype="http://schema.org/Person"><span itemprop="name"><%= link_to_profile @generic_file.depositor %></span></dd>
+  <dt>Date Uploaded</dt>
+  <dd itemprop="datePublished"><%= @generic_file.date_uploaded %></dd>
+  <dt>Date Modified</dt>
+  <dd itemprop="dateModified"><%= @generic_file.date_modified %></dd>
+  <dt>Audit Status</dt>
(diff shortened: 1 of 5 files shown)


app/views/generic_files/_show_details.html.erb
```diff
@@ -0,0 +1,45 @@
...
+  <dd><%= @audit_status %></dd>
+  <% unless @generic_file.related_files.empty? %>
+  <dt>Related Files</dt>
+      <% @generic_file.related_files.each do |f| %>
+          <dd><%= link_to(f.label, sufia.generic_file_url(f)) %></dd>
+      <% end %>
+  <% end %>
+  <dt>Characterization</dt>
+  <dd>
+    <%= "not yet characterized" if characterization_terms.values.flatten.map(&:empty?).reduce(true) { |sum, value| sum && value } %>
```
carolyncole: "characterization_terms.values.flatten.map(&:empty?).reduce(true) { |sum, value| sum && value }" would be a good candidate for a helper method

carolyncole:
@awead  Looking good!  Do you think you would want to deploy this fix on Monday or Tuesday?

carolyncole:
@awead I tried rerunning the travis failure.  It is in the GF show page.  Strange that it is only in the PR.

awead:
@cam156 I updated it with a helper method and additional partial.

PR #152, opened 2015-04-23
Adding a collection partial since that seems to be used by blacklight no...

carolyncole:
...w and not the configuration.  fixes #114


awead:
Should this get added into Sufia at some point? It's not too important.

Issue #164, opened 2015-06-03
solr:compare task always fails


awead:

carolyncole:
https://github.com/psu-stewardship/scholarsphere/commit/ee2e63b8122457ef808d0941b1b4356fed6ca47a

PR #165, opened 2015-06-03
solr:compare task was not working as expected, ref #164

awead:
Although it would have exited properly if solr and fedora were in sync, it would have failed because variables weren't set correctly. This fixes that problem, but it remains to be seen if it reports correctly in our staging and production environments.


carolyncole:
:+1: looks great Adam!

PR #185, opened 2015-09-11
Get ScholarSphere working with the Arkivo API and improve the documentation

mjgiarlo:
We should probably talk this through before merging. I'm not sure the tests pass because this upgrades the stack to the latest in Sufia master (which contains an Arkivo fix). So this Sufia PR needs merging before #185 will work: https://github.com/projecthydra/sufia/pull/1313


mjgiarlo:
The arkivo server segfaults when connecting with Zotero. This fixes me up (it's only a problem on systems with a certain build of gcc): `rm -rf ./node_modules/zotero/node_modules/ws/node_modules/bufferutil`

mjgiarlo:
Talked about this on the 9/14 standup. We're going to let this PR :cool: out until after Hydra Connect, in hopes that Zotero's "My Publications" goes live in production before then (giving us a thing that we can test locally and predictably).

carolyncole:
@mjgiarlo Looks good.  How will we know when My Publications goes online?

mjgiarlo:
@DanCoughlin or I should get contacted, which we can pass on.

Issue #205, opened 2015-10-16
Filter a list of GenericFiles to contain only files that should be sent to share


carolyncole:
Create a service or class that given a list of files can filter that list to only the files we would want to send to Share

The files should be filter by the following:
1. Filter out any file that is not public
2. Pass a file if it has on of the following resource types:
   - Datasets
   - Posters
   - Thesis
   - Dissertation
   - Report
3. Filter out Files already sent to share (this will be based if the file has or does not have a share id)

This is part of the user story #188

PR #207, opened 2015-10-19
Adding filtered lists to allow an array of GenericFiles to be filtere…

carolyncole:
…d based on visibility and resource type refs #205 

These FilteredLists can be used to add filtering need before a file is sen to share.  Implementing them as separate lists so they would be easy to test.

Intending for the final answer to be something like:
send_list =   ResourceFilteredList.new(PublicFilteredList.new(files).filter).filter

There still needs to be an additional filter for not sending ones already sent to Share, but we have yet to define where that data is going to be stored.

### DIFF
--- a/app/models/public_filtered_list.rb
+++ b/app/models/public_filtered_list.rb
@@ -1,0 +1,14 @@
+class PublicFilteredList
+
+  attr_reader :generic_files, :filters
+
+  def initialize(generic_files)
+    @generic_files = generic_files
+    @filters = []
+  end
+
+  def filter
+    @public_members ||= generic_files.reject{|gf| !gf.public?}
+  end
+
+end
--- a/app/models/resource_filtered_list.rb
+++ b/app/models/resource_filtered_list.rb
@@ -1,0 +1,14 @@
+class ResourceFilteredList
+
+  attr_reader :generic_files, :resource_types
+
+  def initialize(generic_files, resource_types = ["Dataset", "Posters", "Thesis", "Dissertation", "Report"] )
+    @generic_files = generic_files
+    @resource_types = resource_types
+  end
+
+  def filter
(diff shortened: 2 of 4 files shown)


app/models/public_filtered_list.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  attr_reader :generic_files, :filters
+
+  def initialize(generic_files)
+    @generic_files = generic_files
+    @filters = []
+  end
+
+  def filter
+    @public_members ||= generic_files.reject{|gf| !gf.public?}
```
awead, line 11: doesn't public_members need an attr_reader?

app/models/resource_filtered_list.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  attr_reader :generic_files, :resource_types
+
+  def initialize(generic_files, resource_types = ["Dataset", "Posters", "Thesis", "Dissertation", "Report"] )
+    @generic_files = generic_files
+    @resource_types = resource_types
+  end
+
+  def filter
+    @filtered ||= generic_files.reject{|gf| !((gf.resource_type & resource_types).count >0)}
```
awead, line 11: do we need an attr_reader for filtered?

app/models/public_filtered_list.rb
carolyncole, line 11: It is just there to allow the call to filter to only apply the filter once.  aka if I run

```
  list = PublicFilteredList.new (file_list)
  puts list.filter
  list.filter.each {|f| do stuff}
```

The above code only runs the filter once based the local variable being set.

app/models/resource_filtered_list.rb
carolyncole, line 11: @awead see the note above for a similar question.  I do not believe we want to allow access directly.

PR #211, opened 2015-11-10
Merging updated solr configs with what's currently in production

awead:
This reconciles the solr configuration that's current in production with what we're using in development. The idea is that the solr configuration in the repo should be identical to what's in production/staging/qa. At the very least, there ought to be only minor variances; however, at this point there's nothing production-specific about the configuration and we can use it in development here without issue.

### DIFF
--- a/solr_conf/conf/elevate.xml
+++ b/solr_conf/conf/elevate.xml
@@ -1,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!--
+ Licensed to the Apache Software Foundation (ASF) under one or more
+ contributor license agreements.  See the NOTICE file distributed with
+ this work for additional information regarding copyright ownership.
+ The ASF licenses this file to You under the Apache License, Version 2.0
+ (the "License"); you may not use this file except in compliance with
+ the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


awead:
ping @cam156 @jrpatterson. Let me know what you think. This would require pushing the files out to QA, Staging and Production. I did some cleanup, so there are only slight differences (mostly formatting) between these files and what's in production. Also, these could be the same configs that are used in ETDA, keep everything more in sync.

solr_conf/conf/elevate.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
carolyncole, line 1: What is this file for?  Is there a need to change it?

solr_conf/conf/schema.xml
```diff
@@ -95,8 +71,6 @@
     <dynamicField name="*_itsim" type="tint" stored="true" indexed="true" multiValued="true"/>
     
     <!-- date (_dt...) -->
-    <!-- The format for this date field is of the form 1995-12-31T23:59:59Z
```
carolyncole, line 98: Having this comment on the date form seems reasonable to me to leave in

solr_conf/conf/elevate.xml
awead, line 1: This is what's currently in production. There needs to be an elevate.xml file for query elevation to work, although it appears to be just a default. See: https://github.com/psu-stewardship/scholarsphere/blob/solr-conf/solr_conf/conf/solrconfig.xml#L1602

solr_conf/conf/schema.xml
awead, line 98: agreed.

solr_conf/conf/elevate.xml
carolyncole, line 1: :ok_hand:

awead:
@cam156  this has been amended and rebased.

PR #227, opened 2016-01-07
Removing Ldp patch, fixes #226

### LINKED ISSUE #226: Remove Ldp::Gone fix
https://github.com/psu-stewardship/scholarsphere/blob/develop/app/models/generic_file.rb#L99-L112
This was put as a "band-aid" and has now been fixed in patch to AF::Noid.

awead:
@cam156 this fixes the issue, and removes the test. I'm not how if we would test this otherwise since I'm not as familiar. Does this look right?


carolyncole:
@awead Looks good!

PR #254, opened 2016-02-19
Refactoring test suite for better effeciency and speed

awead:


lib/devise/behaviors/http_header_authenticatable_behavior.rb
```diff
@@ -9,10 +9,9 @@ def valid_user?(headers)
 
     protected
 
+      # We should only check for HTTP_REMOTE_USER. Web servers will append HTTP_ if it is not present
       def remote_user(headers)
-        return headers['REMOTE_USER'] if headers['REMOTE_USER']
```
carolyncole, line 13: @awead this line was in there because the production server validates the HTTP_REMOTE_USER and passes it along as REMOTE_USER.  I would try deploying this branch to QA and make sure it worked.  Checking REMOTE_USER instead of HTTP_REMOTE_USER was a security feature.

spec/controllers/catalog_controller_spec.rb
```diff
@@ -10,12 +10,7 @@ def contributor_facet
     Solrizer.solr_name("contributor", :facetable)
   end
 
-  let(:user) { FactoryGirl.find_or_create(:user) }
-  before do
-    allow_any_instance_of(GenericFile).to receive(:characterize_if_changed).and_yield
-    sign_in user
-    allow_any_instance_of(User).to receive(:groups).and_return([])
-
+  before(:all) do
```
carolyncole, line 13: @awead Do we need an after all to clean up the files created in the before all?

Interestingly this is closer to what our specs looked like before they were refactored by a contractor.  We went away from the clean up after to the clean before approach.  It seems like the clean up after approach is the fast approach.

spec/controllers/my/files_controller_spec.rb
```diff
@@ -4,41 +4,9 @@
...
-  end
-  before do
-    allow_any_instance_of(User).to receive(:groups).and_return([])
-  end
-  # This doesn't really belong here, but it works for now
-  describe "authenticate!" do
-    before do
-      allow(request).to receive(:headers).and_return('REMOTE_USER' => user.login)
-    end
-    it "populates LDAP attrs if user is new" do
```
carolyncole, line 20: @awead Did these tests get moved elsewhere?

awead, line 20: @cam156 yes, they're in https://github.com/psu-stewardship/scholarsphere/blob/test-refactor/spec/lib/devise/http_header_authenticatable_spec.rb

spec/controllers/catalog_controller_spec.rb
awead, line 13: @cam156 I'm not doing any cleanup before or after the unit tests. We can have rspec do a cleanup before a test if we specify `clean: true` in the test. Because we're not seeing any failures, the cleanup doesn't really have any impact, and **not** cleaning up is how we're seeing such a significant time-savings.

lib/devise/behaviors/http_header_authenticatable_behavior.rb
awead, line 13: Good point. I'll try that now.

spec/features/batch_edit_spec.rb
```diff
@@ -3,110 +3,74 @@
...
 # does not allow us to confirm any updates.
-require_relative './feature_spec_helper'
+require 'feature_spec_helper'
 
 describe 'Batch management of generic files', type: :feature do
-  let!(:current_user) { create :user }
-  let!(:file_1) { create_file current_user, title: 'world.png' }
-  let!(:file_2) { create_file current_user, title: 'little_file.txt' }
-  ##  let(:filenames) { %w{world.png little_file.txt} }
+  let!(:current_user) { FactoryGirl.find_or_create(:user) }
```
carolyncole: @awead this was a problem before, but I do not think we need the let! here.  let should do it.

spec/features/dashboard/dashboard_highlights_spec.rb
```diff
@@ -1,13 +1,13 @@
 # frozen_string_literal: true
-require_relative '../feature_spec_helper'
+require 'feature_spec_helper'
 
 include Selectors::Dashboard
 
 describe 'Dashboard Highlights', type: :feature do
-  let!(:current_user) { create :user }
+  let!(:current_user) { FactoryGirl.find_or_create(:user) }
```
carolyncole: @awead I think the let! is unnecessary here.

spec/features/dashboard/dashboard_spec.rb
```diff
@@ -1,27 +1,28 @@
...
+    let!(:collection) do
+      Collection.new.tap do |c|
+        c.title = "test"
+        c.apply_depositor_metadata(user.user_key)
+        c.save
+      end
+    end
 
     before do
+      sign_in(user)
```
carolyncole, line 25: @awead so this test does not use javascript because we signed in without javascript?

spec/features/dashboard/dashboard_spec.rb
```diff
@@ -1,27 +1,28 @@
 # frozen_string_literal: true
-require_relative '../feature_spec_helper'
+require 'feature_spec_helper'
 
 include Selectors::Dashboard
 
 describe 'The Dashboard', type: :feature do
-  let!(:current_user) { create :user }
+  let(:user) { FactoryGirl.find_or_create(:user) }
```
carolyncole, line 7: @awead curious why the change from current_user to user?  current_user seems more explicitly the logged in user.

spec/features/dashboard/dashboard_spec.rb
awead, line 25: @cam156 yes, it passes either with or without it, but the non-js driver is faster.

spec/features/featured_work_spec.rb
```diff
@@ -1,19 +1,17 @@
 # frozen_string_literal: true
-require_relative './feature_spec_helper'
+require 'feature_spec_helper'
```
carolyncole, line 2: @awead I really like this change.  It looks so much better.

spec/features/dashboard/dashboard_spec.rb
awead, line 7: @cam156 dunno. If you we want it consistent, I guess `current_user` is better.

spec/features/featured_work_spec.rb
awead, line 2: me too! it was really driving me nuts. :laughing:

spec/lib/devise/http_header_authenticatable_behavior_spec.rb
```diff
@@ -1,66 +0,0 @@
-# frozen_string_literal: true
-require 'spec_helper'
```
carolyncole, line 2: @awead why did this entire test get removed?

spec/lib/devise/http_header_authenticatable_spec.rb
```diff
@@ -3,29 +3,39 @@
...
 describe Devise::Strategies::HttpHeaderAuthenticatable do
   subject { described_class.new(nil) }
-  describe "when REMOTE_USER present" do
-    let(:headers) { { "REMOTE_USER" => "abc123" } }
-    before do
-      # I do this in before block or right before test executes
-      @request = double(:request)
-      expect(@request).to receive(:headers).and_return(headers)
-      expect(subject).to receive(:request).and_return(@request)
+  before { allow(subject).to receive(:request).and_return(request) }
```
carolyncole, line 6: @awead Why is this changed from an expect to an allow?  That seems to be changing the flavor of the test.

I wonder if we need to move the expect down into the body of the tests?

spec/models/generic_file_spec.rb
```diff
@@ -3,11 +3,22 @@
 
 describe GenericFile, type: :model do
   let(:file) do
-    described_class.new(id: 'somepid') { |file| file.apply_depositor_metadata('dmc') }
+    described_class.create.tap do |file|
+      file.apply_depositor_metadata('dmc')
+      file.save
```
carolyncole, line 8: @awead Why write to fedora here?  That was not needed before?  That seems slower than just creating an in memory copy.

spec/lib/devise/http_header_authenticatable_behavior_spec.rb
awead, line 2: @cam156 it's all in https://github.com/psu-stewardship/scholarsphere/blob/test-refactor/spec/lib/devise/http_header_authenticatable_spec.rb#L8 although this will change because I discovered we do need to check for REMOTE_USER when I deployed to QA.

spec/lib/devise/http_header_authenticatable_spec.rb
awead, line 6: "allow" goes in a before block. It's in the before block because the test is checking for the presence of the header. How the header got there is not the test's concern.

spec/models/generic_file_spec.rb
awead, line 8: @cam156 it's checking the noid and later, for characterize, which need a saved object.

spec/lib/devise/http_header_authenticatable_behavior_spec.rb
carolyncole, line 2: :ok_hand:

spec/models/generic_file_spec.rb
carolyncole, line 8: @awead but only one of the test is using the noid.  The rest of them are not...

awead, line 8: @cam156 the characterization test fails as well if I don't call `save`.

PR #266, opened 2016-03-15
Refactoring User with LdapUser

awead:
Here are the results of our refactoring sessions, and some additional work I did which mostly extends what we initially started doing.

Comments? @cam156 @hectorcorrea? I don't know where Dan and Derek are on Github.

### DIFF
--- a/app/models/ldap_user.rb
+++ b/app/models/ldap_user.rb
@@ -1,0 +1,66 @@
+# frozen_string_literal: true
+class LdapUser
+  class << self
+    def get_user(filter, fields = [])
+      retry_if { Hydra::LDAP.get_user(filter, fields) } || []
+    end
+
+    def get_groups(login)
+      return [] if login.blank?
+      retry_if { parse_ldap_groups(group_response_from_ldap(login)) } || []
+    end
+
+    def check_ldap_exist!(login)
+      return false if login.blank?
+      retry_if { Hydra::LDAP.does_user_exist?(Net::LDAP::Filter.eq('uid', login)) } || false
+    end
+
+    def filter_for(*people)
+      return "" if people.empty?
+      "(| " + people.map { |p| "(eduPersonPrimaryAffiliation=#{p.to_s.upcase})" }.join(" ") + ")))"
+    end
+
+    private
+
+      def parse_ldap_groups(result)
+        return [] if result.empty?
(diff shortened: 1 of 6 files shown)


app/models/ldap_user.rb
```diff
@@ -0,0 +1,66 @@
...
+    def get_user(filter, fields = [])
+      retry_if { Hydra::LDAP.get_user(filter, fields) } || []
+    end
+
+    def get_groups(login)
+      return [] if login.blank?
+      retry_if { parse_ldap_groups(group_response_from_ldap(login)) } || []
+    end
+
+    def parse_ldap_groups(result)
```
carolyncole: Should this be a private method?

app/models/ldap_user.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+
+    def check_ldap_exist!(login)
+      return false if login.blank?
+      retry_if { Hydra::LDAP.does_user_exist?(Net::LDAP::Filter.eq('uid', login)) } || false
+    end
+
+    # Retries the LDAP command up to .tries times, or catches any other kind of LDAP error without retrying.
+    # return [block or nil]
+    def retry_if
```
carolyncole: Should this be a private method?

carolyncole:
:+1: Looking good!  I'm glad we can remove User from the rubocop ignore now!

app/models/ldap_user.rb
awead: yes, it should.

app/models/ldap_user.rb
awead: yes, it should!

app/models/ldap_user.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+
+    private
+
+      def tries
+        7
+      end
+
+      # Numeric code returned by LDAP if it is feeling "unwilling"
+      def unwilling?
```
hectorcorrea, line 52: It might be a good idea to call this method `ldap_unwilling?` but since it is already inside the `LdapUser` class it might be redundant.

hectorcorrea:
:+1: ship it!

awead:
@hectorcorrea if you're good, I'm good. I've addressed @cam156 comments and it's ready to go.

PR #267, opened 2016-03-16
Use ScholarsphereLockManager for Redis 2.4 compatibility

awead:
This implements a file-based locking mechanism that we use in Archivesphere. I'm testing against Batch, which uses locking in `find_or_create`. It is also used in `Sufia::GenericFile::Actor#add_file_to_collection` but I'm not explicitly testing it there.

The travis build is passing, but it doesn't use Redis 2.4. To test it against 2.4, you'll need to do that locally. I installed 2.4 on my laptop, and the test suite passes except for two errors in `spec/models/migrate_audit_spec.rb`. However, if run that entire spec file in isolation, or even the entire set of model spec tests, they all pass.

Because the failures only pertain to the migration audit, which we're not using anymore, and aren't reproducible, I'm inclined to say this is acceptable and try regression testing on QA.

Thoughts? @cam156 @hectorcorrea @mtribone

### DIFF
--- a/app/services/sufia/lockable.rb
+++ b/app/services/sufia/lockable.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+# Monkeypatch Sufia::Lockable to use a Redis 2.4-compatible lock manager.
+# Remove this when we've upgraded to Redis 2.6+
+module Sufia
+  module Lockable
+    extend ActiveSupport::Concern
+
+    def acquire_lock_for(lock_key, &block)
+      lock_manager.lock(lock_key, &block)
+    end
+
+    def lock_manager
+      @lock_manager ||= ::ScholarsphereLockManager.new
+    end
+  end
+end
--- a/app/services/scholarsphere_lock_manager.rb
+++ b/app/services/scholarsphere_lock_manager.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+# Implements file-based locking which is incompatible with Redis 2.4
+# Remove this when we've upgraded to Redis 2.6+
+class ScholarsphereLockManager
+  def lock(key)
+    state_file = "#{ScholarSphere::Application.config.statefile}_#{key}"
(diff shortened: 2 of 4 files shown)


app/services/sufia/lockable.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
```
hectorcorrea, line 1: Cool, I didn't know about this Ruby pragma.

awead, line 1: Rubocop puts those in there if you target Ruby version 2.3. It's preparation for Ruby 3.0 in which strings will be immutable by default.

hectorcorrea:
What was the error on `spec/models/migrate_audit_spec.rb` ? I agree that we don't need to test the migration per-se but I wonder if it really uncovered an error during `GenericFile.create`.

awead:
``` bash

  1) MigrateAudit with a file migrated correctly returns OK
     Failure/Error: expect(subject.status).to eq("OK")

       expected: "OK"
            got: "Not found"

       (compared using ==)
     # ./spec/models/migrate_audit_spec.rb:31:in `block (3 levels) in <top (required)>'


  2) MigrateAudit with a file migrated incorrectly reports models mismatch
     Failure/Error: expect(subject.status).to eq("Models mismatch. Expected: GenericFileFake but found: GenericFile")

       expected: "Models mismatch. Expected: GenericFileFake but found: GenericFile"
            got: "Not found"

       (compared using ==)
     # ./spec/models/migrate_audit_spec.rb:48:in `block (3 levels) in <top (required)>'
```

That's a good point. I guess we'll find out when we test it on QA.

awead:
@cam156 @hectorcorrea So I pushed this up to QA and uploaded a couple of batches of files, then added those files to collections. All went well. I could run through the whole battery of tests, or otherwise move to a scheduled release and run the full tests then with this PR merged in develop.

hectorcorrea:
@awead sounds good. We could even add a specific test to the Wiki to highlight the need to test this heavily for the next release or two.

PR #268, opened 2016-03-29
Delete documents from SHARE Notify

awead:
Supports deleting documents from SHARE Notify if the file either becomes private, or is deleted from Scholarsphere.


awead:
ping @cam156 @hectorcorrea

app/controllers/generic_files_controller.rb
```diff
@@ -16,15 +15,24 @@ def new
     @batch_id = Batch.create.id
   end
 
-  around_action :notify_users_of_permission_changes, only: [:destroy, :create, :update]
+  around_action :notify_users_of_permission_changes, only: [:update]
```
carolyncole, line 18: @awead Why are you removing the notification on create? I'm assuming the notification for delete moved?

app/models/permissions_change_set.rb
```diff
@@ -0,0 +1,55 @@
...
+#        removed: [ { name: "xyz123", type: "group", access: "read" }, ... ]
+#      }
+#
+class PermissionsChangeSet
+  attr_reader :before, :after
+
+  # @param [Hash] state of a file's permissions before a change
+  # @param [Hash] state of a file's permissions after a change
+  def initialize(*args)
+    @before, @after = args
```
carolyncole: @awead I'm not sure I am in :heart: with this syntax.  Sure it save a line, but...

app/services/permissions_change_service.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    def delete_from_share
+      Sufia.queue.push(ShareNotifyDeleteJob.new(generic_file.id))
+    end
+
+    def send_message(access, recipient = nil)
+      return if recipient.nil?
+      User.batchuser.send_message(
+        recipient,
+        "You can now #{access} file #{generic_file.title}",
```
carolyncole, line 50: @awead why remove the I18N?

spec/controllers/generic_files_controller_spec.rb
```diff
@@ -4,13 +4,7 @@
...
   routes { Sufia::Engine.routes }
   context "public user" do
-    let(:gf) {
-      GenericFile.new(title: ['Test Document PDF'], filename: ['test.pdf'], read_groups: ['public']).tap do |gf|
-        gf.apply_depositor_metadata("mjg36")
-        gf.save!
-      end
-    }
-
+    let(:gf) { create(:public_file) }
```
carolyncole, line 7: @awead :+1: this looks much better!

app/controllers/generic_files_controller.rb
awead, line 18: @cam156 If you look at the original `notify_users` method, which is now removed, it was only executing on update. We could put this back in, but it will need to change significantly since the file event notification cannot be made if the file is deleted. For creation, I'm not sure. We'd need to add additional test coverage. If we want to put this in, I can look into it via another issue.

app/models/permissions_change_set.rb
awead: If there's reason not to use it, I'm open to changing it.

app/services/permissions_change_service.rb
awead, line 50: It simplifies things, and keeps things consistent. In Sufia 7, MessageUserService does not use I18N and defines the content of the message in the class.

app/controllers/generic_files_controller.rb
carolyncole, line 18: :ok_hand:

app/models/permissions_change_set.rb
carolyncole: @awead it just seems more ambiguous than just accepting 2 arguments to the constructor.  Why say we accept a hash when we really want two specific items

app/services/permissions_change_service.rb
carolyncole, line 50: @awead with how much the wording is likely to change between users I am surprised that Sufia is heading down that path. I'm :ok: with it here, but it does go against the general push to have text convertible.

app/controllers/generic_files_controller.rb
awead, line 18: @cam156 I've added issue #269

app/models/permissions_change_set.rb
awead: `args` is an array of two hashes. If we want a tighter interface then we can just do:

``` ruby
def initialize(before, after)
  @before = before
  @after = after
end
```

In general, it's pretty forgiving. I just elected to use it as a form of shorthand.

``` ruby
irb(main):001:0> a, b = "a", "b"
=> ["a", "b"]
irb(main):002:0> a
=> "a"
irb(main):003:0> b
=> "b"
irb(main):004:0> a, b = nil, "b"
=> [nil, "b"]
irb(main):005:0> a
=> nil
irb(main):006:0> b
=> "b"
irb(main):007:0> a, b = "a"
=> "a"
irb(main):008:0> a
=> "a"
irb(main):009:0> b
=> nil
irb(main):010:0> a, b = nil, "b", "foo"
=> [nil, "b", "foo"]
irb(main):011:0> a
=> nil
irb(main):013:0> b
=> "b"
```

app/services/permissions_change_service.rb
awead, line 50: I'd be interesting in exploring that along with our jobs. They have all their text defined in the class as well.

awead:
@cam156 this is ready to go.

PR #271, opened 2016-04-01
Use factories in our tests

awead:


awead:
ping @cam156 @hectorcorrea

spec/controllers/catalog_controller_spec.rb
```diff
@@ -139,13 +106,13 @@ def contributor_facet
         expect(assigns(:document_list).count).to eql(1)
       end
       it "finds a file by depositor" do
-        xhr :get, :index, q: "mjg36"
+        xhr :get, :index, q: "user"
```
carolyncole: @awead This seems like it might fail at some point since you are assuming text that is not in the current file.  It is a magic string.
possibly change it to current_user.id or public file.depositor?

spec/controllers/directory_controller_spec.rb
```diff
@@ -3,7 +3,7 @@
 
 describe DirectoryController, type: :controller do
   routes { Sufia::Engine.routes }
-  let(:user) { FactoryGirl.find_or_create(:user) }
+  let(:user) { create(:user) }
```
carolyncole, line 6: @awead what is the difference between create and build?  If there is no difference maybe we should pick a preferred method.

spec/controllers/my/files_controller_spec.rb
```diff
@@ -53,29 +43,7 @@
...
-          f.subject = ["subjectsubject"]
-          f.resource_type = ["resource_typeresource_type"]
-          f.resource_type = ["resource_typeresource_type"]
-          f.description = ["descriptiondescription"]
-          f.format_label = ["format_labelformat_label"]
-          f.apply_depositor_metadata(user.login)
-          f.update_index
-        end
-      end
+      before(:all) { create(:public_file, :with_complete_metadata, depositor: "archivist1") }
```
carolyncole, line 46: @awead :heart: the trait!

spec/factories/files.rb
```diff
@@ -1,10 +1,19 @@
 # frozen_string_literal: true
 FactoryGirl.define do
   factory :file, aliases: [:private_file], class: GenericFile do
+    ignore do
```
carolyncole, line 4: @awead what does this do?

spec/factories/files.rb
```diff
@@ -15,10 +24,60 @@
         creator ["Joe Contributor"]
         resource_type ["Dissertation"]
       end
+
+      factory :featured_file do
```
carolyncole, line 28: @awead These factories may be hard to find since they are hidden.  I'm ok with the location, but another time a separate file might be better.

spec/features/collection/edit_spec.rb
```diff
@@ -4,107 +4,86 @@
...
+    let!(:file3)      { create(:file, title: ['scholarsphere_test5.txt'], depositor: current_user.login) }
 
-  describe 'When adding a file to a collection' do
-    before do
-      db_file_checkbox(file_3).click
-      click_button 'Add to Collection'
-      db_collection_radio_button(collection).click
-      within("#collection-list-container .modal-footer") do
+    describe 'adding an additional file' do
+      specify do
```
carolyncole, line 19: @awead not sure I love the specify with no text here.  What are we specifying?

spec/controllers/catalog_controller_spec.rb
awead: It's the default user value specified in:
https://github.com/psu-stewardship/scholarsphere/blob/develop/spec/factories/files.rb#L6
This is to save you from having to specify a depositor for a file if it's not important.

In the context of this test, there's no user logged in, so we can't use `current_user`, but could either create a user and make them the depositor of the files, or have a `deafult_user` method that returns "user" for the test and the factory.

Since we're needing to search on the username, perhaps just specifying a user after all?

spec/controllers/directory_controller_spec.rb
awead, line 6: It's the same as `new` versus `create`. Calling `build` won't persist the object, and is preferable since it's faster. But I think in this case, we needed a persisted user because we're searching for it somewhere in this test.

spec/factories/files.rb
awead, line 4: This is so you can pass arbitrary methods such as `transer_to`. See:
https://github.com/thoughtbot/factory_girl/blob/v4.4.0/GETTING_STARTED.md#transient-attributes

awead, line 4: ignore has been deprecated, but I was getting NoMethodError when trying to use transient.

spec/factories/files.rb
awead, line 28: Yeah, the nesting makes them harder to see. Another alternative is:

``` ruby
FactoryGirl.define do
  ...
  factory :featured_file, parent: public_file do
    ...
  end
end
```

That would move the block over to the left more, otherwise, I'm fine with another file.

spec/features/collection/edit_spec.rb
awead, line 19: It's in the describe block. Previously, we had something like:

``` ruby
describe "adding an additional file" do
  it "adds an additional file"
    ...
  end
end
```

which seemed redundant to me. I reserve `specify` for cases in which there isn't a description needed. It's purely subjective.

spec/controllers/directory_controller_spec.rb
carolyncole, line 6: :cool:

spec/factories/files.rb
carolyncole, line 28: @awead lets leave them for now and move them later if we find we are changing them a lot.

spec/features/collection/edit_spec.rb
carolyncole, line 19: :ok_hand:

carolyncole:
@awead once this is :green_heart: I will merge...

PR #272, opened 2016-04-04
Export s6

hectorcorrea:
*\* DO NOT MERGE **

This is a first pass at exporting ScholarSphere's GenericFiles and Collections stored under a Sufia 6 implementation to a set of JSON files that can be used to import the data into a Sufia 7 implementation. 

To export Generic Files the idea is to run two rake tasks. One to export the IDs of all the files to export and another to actually export them to JSON files (one per Generic File). I suspect we could combine these tasks but for a final release. 

```
rake gf_ids > gf_ids.txt
rake export_gf[gf_ids.txt]
```

Similar rake tasks can be used to export Collections.

You can see a sample of the JSON files generated when exporting a file and a collection here: https://gist.github.com/hectorcorrea/619c13fa540341be14cb01c762d9f2cc

The JSON file generated for Generic Files only includes the metadata (including metadata for versions and permissions) but not the actual binary. The idea is to fetch the binary (for the current and previous versions) directly from Fedora 4 at the time the data is imported to a Sufia 7 application.

### DIFF
--- a/lib/export/collection_metadata_export.rb
+++ b/lib/export/collection_metadata_export.rb
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
+
+# This class is used to consolidate Collection metadata in a way that
+# can be easily exported to JSON.
 module Export
-  class CollectionExport
+  class CollectionMetadataExport
     attr_accessor :id, :title, :depositor, :description, :creator, :members
 
     def initialize(coll)
--- a/lib/export/generic_file_metadata_export.rb
+++ b/lib/export/generic_file_metadata_export.rb
@@ -1,9 +1,12 @@
 # frozen_string_literal: true
-require "./lib/export/permissions.rb"
+require "./lib/export/permissions_metadata.rb"
 require "./lib/export/versions_from_graph.rb"
 
+# This class is used to consolidate GenericFile metadata in a way that
+# can be easily exported to JSON. Notice that we don't store the
+# actual binary of the file here.
 module Export
-  class GenericFileExport
+  class GenericFileMetadataExport
     # Properties to be exported
     attr_accessor :id, :label, :depositor, :arkivo_checksum, :relative_path,
                   :import_url, :resource_type, :title, :creator, :contributor,
@@ -44,7 +47,7 @@ module Export
       if gf.content.has_versions?
         @versions = Export::VersionsFromGraph.parse(gf.content.versions)
       end
-      @permissions = Export::Permissions.new(gf.permissions).to_a
+      @permissions = Export::PermissionsMetadata.new(gf.permissions).to_a
     end
 
     def to_json(pretty = false)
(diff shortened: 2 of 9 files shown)


lib/export/generic_file_metadata_export.rb
```diff
@@ -0,0 +1,58 @@
...
+      @date_modified = gf.date_modified
+      @subject = gf.subject
+      @language = gf.language
+      @identifier = gf.identifier
+      @based_near = gf.based_near
+      @related_url = gf.related_url
+      @bibliographic_citation = gf.bibliographic_citation
+      @source = gf.source
+      @batch_id = gf.batch.id if gf.batch
+      @visibility = gf.visibility
```
awead, line 45: use delegate_to for all of these :smile:

lib/export/versions_from_graph.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+
+    def self.find_uris(graph)
+      uris = []
+      graph.each do |triple|
+        if triple.predicate.to_s == "http://fedora.info/definitions/v4/repository#hasVersion"
+          uris << triple.object.to_s
+        end
+      end
+      uris
```
awead, line 35: I think you can do something like:

``` ruby
graph.query(predicate: ::RDF::Vocab::Fcrepo4.hasVersion).map(&:to_s)
```

lib/export/versions_from_graph.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+module Export
+  class VersionsFromGraph
+    attr_accessor :versions
```
awead, line 4: is `ActiveFedora::VersionsGraph` any help here? Not sure, just wanted to make sure you knew about it.

lib/export/generic_file_metadata_export.rb
```diff
@@ -0,0 +1,58 @@
...
+      @date_uploaded = gf.date_uploaded
+      @date_modified = gf.date_modified
+      @subject = gf.subject
+      @language = gf.language
+      @identifier = gf.identifier
+      @based_near = gf.based_near
+      @related_url = gf.related_url
+      @bibliographic_citation = gf.bibliographic_citation
+      @source = gf.source
+      @batch_id = gf.batch.id if gf.batch
```
carolyncole, line 44: Change this to list the related IDs from the original batch inside the Item.

lib/export/collection_metadata_export.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    def initialize(coll)
+      @id = coll.id
+      @title = coll.title
+      @description = coll.description
+      @creator = coll.creator.map { |c| c }
+      @members = coll.members.map(&:id)
+    end
+
+    def to_json(pretty = false)
```
carolyncole, line 17: Overrode so that it is easier to debug.  This is not needed for running.

PR #292, opened 2016-05-12
Collection edit test passing

carolyncole:
collection/edit test pass
And fixed up the error created when Mona left the university

### DIFF
--- a/spec/features/collection/edit_spec.rb
+++ b/spec/features/collection/edit_spec.rb
@@ -12,7 +12,7 @@ describe Collection, type: :feature do
   before { sign_in_with_js(current_user) }
 
   context "when the collection has files" do
-    let!(:collection) { create(:collection, depositor: current_user.login, members: [work1, work2]) }
+    let!(:collection) { create(:collection, depositor: current_user.login, members: [work1, work2], description: ["my description"]) }
     let!(:work3)      { create(:work, title: ['scholarsphere_test5.txt'], depositor: current_user.login) }
 
     describe 'adding an additional file' do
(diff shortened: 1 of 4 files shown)


spec/features/collection/edit_spec.rb
```diff
@@ -52,35 +52,36 @@
...
+    # let!(:original_title)       { collection.title }
+    # let!(:original_description) { collection.description }
+    #
+    # let(:updated_title)         { 'Updated Title' }
+    # let(:updated_description)   { 'Updtaed description text.' }
+    # let(:updated_creators)      { ['Dorje Trollo', 'Vajrayogini'] }
 
     specify do
+      visit '/dashboard/highlights'
       visit '/dashboard/collections'
```
hectorcorrea, line 73: Do we need to visit both pages (highlights + collections) ?

carolyncole:
@hectorcorrea I think this one is a merge of our changes now...

spec/features/collection/edit_spec.rb
```diff
@@ -12,7 +12,7 @@
   before { sign_in_with_js(current_user) }
 
   context "when the collection has files" do
-    let!(:collection) { create(:collection, depositor: current_user.login, members: [work1, work2]) }
+    let!(:collection) { create(:collection, depositor: current_user.login, members: [work1, work2], description: ["my description"]) }
```
hectorcorrea, line 15: You shouldn't need to set the description here, but it won't cause any harm either.

carolyncole, line 15: @hectorcorrea since we are checking it I figure it is more explicit to set it.  The change in the factory makes sure that the description doesn't cause harm anywhere...

hectorcorrea:
@cam156 Could you rebase it?

carolyncole:
@hectorcorrea should be good now.

PR #309, opened 2016-05-24
Getting GenericFile show and download spec to work for Sufia 7

carolyncole:


hectorcorrea:
@cam156 Rubocop is not happy.

carolyncole:
@hectorcorrea rubocop is happier now!

PR #327, opened 2016-07-22
getting no javascript errors

carolyncole:


spec/features/static_pages_spec.rb
```diff
@@ -30,7 +30,7 @@
 
   context 'when not logged in' do
     before do
-      sign_in_with_js(nil, js_errors: false)
+      sign_in_with_named_js(:help_poltergeist, nil, js_errors: false)
```
awead, line 33: Is this because you need to instantiate a new poltergeist driver?

carolyncole, line 33: @awead yes.  If we use the old name a new driver does not get created and the configuration stays the same.

PR #329, opened 2016-08-01
Updates SS for latest brand guidelines and tweaks styles for wider mark.

mtribone:
screen shot 2016-08-01 at 3 44 31 pm

screen shot 2016-08-01 at 3 45 10 pm

screen shot 2016-08-01 at 3 45 40 pm


kestlund:
Is it possible to make "Scholar Sphere" a bit more prominent / bolder within the new guidelines?  Or is this the limit?

Thanks for the update @mtribone !

mtribone:
@kestlund I might be able to bump the "ScholarSphere" a wee bit on the larger width and see how that goes with the Comms group. It's not supposed to "compete" with the University mark, but we'll give it a shot!

olendorf:
I agree with increasing the "prominence" of scholarsphere as much as possible as well. It gets kinda lost as is.

app/assets/stylesheets/_header-footer.scss
```diff
@@ -1,67 +1,56 @@
...
-    color: $classic-white;
-    display: block;
-    font-family: 'Lato', Verdana, Arail, Helvetica, sans-serif;
-    font-size: 2em;
-    font-weight: 300;
-    text-decoration: none;
-    padding-top: .5em;
-  }
-  .logoBold {
-    font-weight: 500;
```
carolyncole, line 17: I think this removal may be what is making the ScholarSphere look smaller.  This class used to be applied by default.

mtribone, line 17: Yes. I intentionally refactored the code and made the minimum size smaller per the branding guidelines. The code that used to be outside of the media query has been refactored so that the starting point is for mobile.

https://github.com/psu-stewardship/scholarsphere/blob/e31eeb4caa6d7575507f642ff560a368e324eb9e/app/assets/stylesheets/_header-footer.scss#L15

app/assets/stylesheets/_header-footer.scss
```diff
@@ -1,67 +1,56 @@
...
   }
 }
 
-@media screen and (min-width: 781px) {
+@media screen and (min-width: 785px) {
   #masthead {
     .logo a:link, .logo a:visited, .logo a:hover {
-      font-size: 3em;
-      padding-top: .15em;
+      font-size: 2em;
```
mtribone, line 41: I'll target this screen size and bump the font size a bit and see if IT Comms calls "foul."

mtribone:
Alright. I've bumped the size a bit when the viewport is at a minimum of 912 pixels. An issue with anything that is smaller than this and larger font sizes are the number of Bootstrap columns for the logo section and the user section. It isn't reflected in the screen shot, but that button combo on the right will need to accommodate _full_ names from ldap.

screen shot 2016-08-02 at 3 30 57 pm

mtribone:
@kestlund @olendorf I got the "looks good" from the Comms group for the above ⏫  change. Let's put it on staging and see how it looks in a responsive setting. Sound good?

mtribone:
We could also use a different color for "ScholarSphere"

awead:
@mtribone  Looks great. Could you sqaush your commits and/or fix the typo in the commit message?

mtribone:
_golf claps_

Issue #330, opened 2016-08-04
Penn State Mark, ScholarSphere, Column Tweaks (400px to 768px)


mtribone:
I think that there might be some room for improvement with the number of columns and size of the ScholarSphere text, since we don't have a full user bar with the full name until 768px.

Challenges
1. col-xs  is < 768 pixels
2. col-sm is >= 768 pixels

Possible solutions
1. Reduce size of Penn State mark within a certain range (400...600)
2. Increase size of ScholarSphere within a certain range (500...700)

Extras
1. Text color for ScholarSphere
2. Letterspacing

![screen shot 2016-08-04 at 9 23 27 am](https://cloud.githubusercontent.com/assets/4163828/17403342/474e0558-5a25-11e6-8127-0c7808346b1a.png)

![screen shot 2016-08-04 at 9 12 47 am](https://cloud.githubusercontent.com/assets/4163828/17403249/eb4b3258-5a24-11e6-92a0-05e568db1a8a.png)

carolyncole:
#328 will fix this

PR #339, opened 2016-08-04
Responsive Penn State mark and larger ScholarSphere based on browser width.

mtribone:
So I decided to slice it a little finer and really dictate what the sizes are for things based on browser width up to 1024 pixels. If we need moar logo past that number we'll need to tweak some more.

Refs #330

### DIFF
--- a/app/assets/stylesheets/_header-footer.scss
+++ b/app/assets/stylesheets/_header-footer.scss
@@ -1,8 +1,9 @@
-@media screen and (min-width: 400px) {
+@media screen and (min-width: 350px) {
   #masthead {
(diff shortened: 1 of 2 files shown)


mtribone:
400px

screen shot 2016-08-04 at 3 12 05 pm

mtribone:
650px

screen shot 2016-08-04 at 3 13 55 pm

mtribone:
800px

screen shot 2016-08-04 at 3 15 47 pm

mtribone:
1024px

screen shot 2016-08-04 at 3 17 03 pm

PR #356, opened 2016-08-11
Configuring to use resque

### LINKED ISSUE #333: Make Sure ScholarSphere is configured for background jobs
It seems like the jobs are being run inline.

We need these jobs to be run in the background so the system seems more responsive.

?? How do we tell if this is complete??
Disable resque, run the upload, wait and see not thumbs generate, enable resque and see thumbs?

carolyncole:
fixes #333

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -8,6 +8,9 @@ ScholarSphere::Application.configure do
   # Settings specified here will take precedence over those in config/application.rb
   config.log_level = :debug
 
+  # set up resque in production only (uncomment if you would like reque locally)
+  # config.active_job.queue_adapter = :resque
+
   # In the development environment your application's code is reloaded on
   # every request.  This slows down response time but is perfect for development
   # since you don't have to restart the webserver when you make code changes.
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -7,6 +7,9 @@ end
 ScholarSphere::Application.configure do
   # Settings specified here will take precedence over those in config/application.rb
 
+  # set up resque in production only
+  config.active_job.queue_adapter = :resque
+
   # Set cookies as secure, force all connections over SSL
   config.force_ssl = true
 


awead:
@cam156 one thought: run them inline for development and test and we see failures in the job chain immediately, as opposed to having to find them in resque? Of course, if there's a problem with resque we wouldn't see it until we deployed.

carolyncole:
@awead I'm +-0 on running them inline in development.  I agree with both your statements.  It would be easier to see failures, but we would not see resque timing issues ect. until deployment.  I usually lean toward having the development environment as close to production as possible, but I could go either way... @mtribone have any opinion?

mtribone:
I want my 🍰 and 😮 , too. Soooo if we run them inline in dev, the downside is that we need to deploy to another server to get a sense of resque timing issues. I think that I'm good with it. We can see failures as we develop, we just need to test timing outside of dev–which might be fine since local machine != remote server.

awead:
FWIW, I share @cam156's concerns too... I was just floating the idea. We can try it out for now and see how it goes.

mtribone:
Is this where we need the voice of @hectorcorrea  to say, "Well...wait a minute."

awead:
hehe. maybe! The nice thing about this option is that when you create a work, all the jobs are run inline and the results get written to log/development.log. It's nice to see all that in one place.

OTOH, you can easily change this locally by just commenting-out the one line in application/config.rb

carolyncole:
👌 I'll modify the PR to use the configuration only in Production.

carolyncole:
@mtribone @awead I moved where the configuration was occurring and put a commented out version in development incase someone wants to enable it.  It seems like a good compromise.

awead:
@cam156 awesome. I'll merge this as soon as Travis is green.

hectorcorrea:
Rubocop needs to get a life ;)

PR #357, opened 2016-08-11
Adding Questioning Authority routes and configuration sample.

### LINKED ISSUE #332: Need to be certain the Autocomplete for location works
Need to be certain the autocomplete for location is configured properly.  It is currently not working on my local server or on QA so the configuration must have changes in Sufia 7.

This is complete when upload a new work or edit an existing work and get an autocomplete on the Location field

carolyncole:
Fixes  #332

The actual username is documented in sites.

### DIFF
--- a/config/initializers/sufia-secret.rb.sample
+++ b/config/initializers/sufia-secret.rb.sample
@@ -1,0 +1,4 @@
+Sufia.config do |config|
+    config.geonames_username = 'username'
+end
+
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -61,7 +61,8 @@ set :linked_files, fetch(:linked_files, []).push(
   'config/secrets.yml',
   'public/sitemap.xml',
   'public/robots.txt',
-  'config/initializers/arkivo_constraint.rb'
+  'config/initializers/arkivo_constraint.rb',
+  'config/initializers/sufia-secret.rb'
 )
 
 set :linked_dirs, fetch(:linked_dirs, []).push(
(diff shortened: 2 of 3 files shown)


config/initializers/sufia-secret.rb.sample
```diff
@@ -0,0 +1,4 @@
+Sufia.config do |config|
+    config.geonames_username = 'username'
+end
+
```
awead, line 4: How about putting this `config/intializers/sufia.rb` and defining it via `config/scholarsphere.yml` ? We can manage it using Capistrano after that.

carolyncole: @awead We get limited by the number of calls based on our user name.  If we put that into an open profile someone could user our user name and we could get limited by geo users.  I think it is much better to keep the user name our own, which is why I chose to put it in secrets.

awead: @cam156 if we manage scholarsphere.yml by capistrano, the repo file would be overridden on deploy.

carolyncole: @awead is scholarsphere.yml already managed via cap?

carolyncole: @awead good idea to have it managed by cap. From what I can tell scholarSphere.yml does not exists and is not managed by cap. Am I missing something?  Why in a yml instead of in a .rb as sufia suggest/requires?

awead: @cam156 it can be. We could also do it with sufia.rb too. Any file, as far as I know, can be handled this way. All you do is add it to the list of files in deploy.rb. We would need to add it on the server though. There is one drawback here that I didn't realize. If you change the configuration locally in your repo to use your own username, you won't be able to save it because it would get committed. So maybe the sample file is the better option.

carolyncole:
@awead how about this.  Leave it as a sample and add a line for the real thing in cap?  Work for you?  We already had a sufia-secrets.rb out on the servers, but it was no longer being used.  I updated it to include the correct information of QA

PR #372, opened 2016-09-09
Using the latest Sufia from master

awead:
There will be some, or maybe many, failures here, so this is a question. Do we want to roll with the latest and greatest in Sufia for our upcoming sprint? 7.2 should be released in a week or less, so it's likely it'll come out during the sprint. So I'm think we should just start working against it now to minimize surprises.


awead:
@cam156 this is ready to go. I've pushed up the latest commit from Sufia as of today. I've marked one test pending because of a peculiar failure that I've outlined in https://github.com/psu-stewardship/scholarsphere/pull/372.

config/authorities/licenses.yml
```diff
@@ -0,0 +1,28 @@
+terms:
```
carolyncole, line 1: @awead I think we should just rename rights.yml instead of adding a new file.

spec/features/batch_edit_spec.rb
```diff
@@ -19,7 +19,7 @@
       work1.reload
       work2.reload
     end
-    it 'edits each field and displays the changes', js: true do
+    xit 'edits each field and displays the changes', js: true do
```
carolyncole, line 22: @awead Is this the intentionally pending test?

spec/presenters/stats_presenter_spec.rb
```diff
@@ -72,10 +72,10 @@
   end
 
   describe "#total_users" do
-    let(:start_datetime) { DateTime.parse("2004-01-01T01:01:01") }
-    it "calls Sufia::Statistics::SystemStats for data" do
-      expect(system_stats).to receive(:users_count).and_return(100)
-      expect(presenter.total_users).to eq(100)
+    let(:users_count)    { ::User.count }
```
carolyncole: This seems like it might be testing zero.  You should create a user to make sure it does not always return zero.

carolyncole (changes requested):
Looks good overall.  Seems like you have 3 new tests that are pending.  One that you changes the it to xit, one marked pending and one that was commented out that is now pending.

config/authorities/licenses.yml
awead: It was changed in Sufia. I can just remove the rights.yml file.

spec/features/batch_edit_spec.rb
awead: Yes, batch edit is misbehaving as well.

carolyncole (approved):

PR #395, opened 2016-10-11
Reconfiguring test suite to use Rspec rails_helper file

awead:
By default, new Rspec installations within a Rails application create a spec_helper file and a rails_helper file. Errors can arise if the latter is not present. Additionally, the updated spec_helper file enables `verify_partial_doubles` which caused new failures because we were mocking methods that no longer existed on the test subject. The tests have been corrected, and now with the newly-enabled option, if we mock methods that aren't present in the test subject, the test will fail.

FYI, @cam156 @mtribone


spec/views/catalog/_index_default.html.erb_spec.rb
```diff
@@ -1,35 +0,0 @@
-require 'spec_helper'
-
-describe "catalog/_index_default.html.erb" do
```
carolyncole, line 3: @awead Why are you removing this test?

spec/controllers/catalog_controller_spec.rb
```diff
@@ -13,7 +13,6 @@
   let(:file) { mock_file_factory(format_label: ['format_labelformat_label']) }
 
   before do
-    allow_any_instance_of(GenericWork).to receive(:characterize_if_changed).and_yield
```
carolyncole, line 16: @awead Curious why you removed this?

carolyncole (changes requested):
Overall looks great!  Not sure why an entire test got removed.

spec/controllers/catalog_controller_spec.rb
awead: This is what I talking about by enabling `verify_partial_doubles`. The `characterize_if_changed` doesn't exist and now the test fails because we're mocking a method that doesn't exist anymore.

spec/views/catalog/_index_default.html.erb_spec.rb
awead: Because of `verify_partial_doubles`, the test fails because that view doesn't implement `blacklight_configuration_context`. I believe this is because we're not directly overriding this view. In any case, the test criteria was moved to spec/features/catalog_spec.rb. Incidentally, the view test was passing when it should have been failing because the titles do in fact appear in the search results when in gallery view, and the test was asserting that they did not.

carolyncole (approved):

PR #426, opened 2016-10-26
Fluid container and spacing

mtribone:
Fixes container on home page and 1 column layout from CC to be container-fluid.

### DIFF
--- a/app/assets/javascripts/scholarsphere_layout.js
+++ b/app/assets/javascripts/scholarsphere_layout.js
@@ -1,0 +1,7 @@
+/**
+ * Add/remove a class to/from a DIV and not override a whole view
+ */
+
+Blacklight.onLoad(function() {
+    $( "#content-wrapper" ).removeClass( "container" ).addClass( "container-fluid" );
+});


mtribone:
screen shot 2016-10-26 at 10 25 21 am

mtribone:
screen shot 2016-10-26 at 10 25 42 am

carolyncole (commented):
Is this change just about adding "container-fluid" to the content-wrapper div?

Could we just add that via javascript?

```
document.getElementById('content-wrapper').classList.add('container-fluid');
```

Or add some TODOs at the beginning of the class with a ticket in sufia so we can stop maintaining this code at some point.

mtribone:
We certainly could just add it via javascript. I'd have to remove the class `container` and add the class `container-fluid`

app/assets/javascripts/scholarsphere_layout.js
```diff
@@ -0,0 +1,5 @@
+/**
```
carolyncole, line 1: @mtribone You likely need the BlackLight.onload like here: https://github.com/projecthydra/sufia/blob/master/app/assets/javascripts/sufia/batch.js#L1
and then close like here: https://github.com/projecthydra/sufia/blob/master/app/assets/javascripts/sufia/batch.js#L13

app/assets/javascripts/scholarsphere_layout.js
```diff
@@ -0,0 +1,5 @@
...
+ * Add/remove a class to/from a DIV and not override a whole view
+ */
+
+document.getElementById('content-wrapper').classList.add('container-fluid');
```
carolyncole: You can also remove a class.  Here is the jquery syntax for adding and removing a class

`$( "#content-wrapper" ).removeClass( "container" ).addClass( "container-fluid" );`

So the entire file would look more like:

```
Blacklight.onLoad(function() {
   $( "#content-wrapper" ).removeClass( "container" ).addClass( "container-fluid" );
});
```

carolyncole (changes requested):

app/assets/javascripts/scholarsphere_layout.js
awead: changes approved!

PR #432, opened 2016-11-01
Moved User Profile Info from Left-Hand Column into Tab Pane

mtribone:
Moves the user info from the left pane and into its own tab panel. Removes extra col-xs-12 and sets up left sidebar and tab content to behave and display more consistently on mobile and up.

Swaps the active tab to the user info and adds header information to each column, adds h1 to page and hides it visually for screen readers, makes user name h2, removes the extra edit button, and updates spec test for the correct link text.

Adds I18n translation for User Info tab. Font size and color tweaks.

Fix up the table a bit for style, display, and accessibility.

### DIFF
--- a/app/assets/stylesheets/scholarsphere/user_profile.scss
+++ b/app/assets/stylesheets/scholarsphere/user_profile.scss
@@ -1,0 +1,14 @@
+.highlighted-works {
+  margin-top: 1em;
+}
+
+.highlighted-works td {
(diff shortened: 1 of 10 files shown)


mtribone:
screen shot 2016-11-01 at 1 51 18 pm

mtribone:
screen shot 2016-11-01 at 1 51 31 pm

PR #452, opened 2016-11-14
Keyboard Navigation and Associated Accessibility Work for New/Edit Work Interface

mtribone:
Makes `<span />` a `<label />` and updates styles for padding, font weight, and increasing the clickable area of the button.

Moves the main label for embargo and lease to surround just those elements and not around a huge nested set of inputs. Adds values to the add buttons for better usability and removes several sr-only to provide instructions for all users in the interface.

Unhides help text, puts it into an inline ordered list, and adds the help block styles. Removes a label element from a span element for better accessibility and reworks the sharing table to be a bit easier to read and understand for users with differing abilities.

Adjust the display of the select buttons and make room for the remove button. Wordsmith help text. Refactors the sharing table abit and adds labels to the date picker on the new work form. Removes some extra parameters which weren't really working and adds label defaults to sufia yml.

Remove the bootstrap btn and btn-success to add own styles for tabbed focus. Fixes up the file names and IDs plus CSS for better alignment. Adds jquery event listeners for tabbed focus. Updates border, box shadow and outline to match system highlighting.

### DIFF
--- a/app/assets/javascripts/scholarsphere_fileupload.js
+++ b/app/assets/javascripts/scholarsphere_fileupload.js
@@ -5,3 +5,11 @@ max_file_size_str = "500 MB";
 max_total_file_size = 1000000000;
 max_total_file_size_str = "1000 MB";
 
+Blacklight.onLoad(function() {
+    $(":input").focusin(function () {
+        $(this).addClass("has-focus");
+    });
+    $(":input").focusout(function () {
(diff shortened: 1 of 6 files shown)


mtribone:
screen shot 2016-11-14 at 10 24 00 am

mtribone:
screen shot 2016-11-14 at 10 24 49 am

PR #489, opened 2016-12-07
[WIP] Hide thumbnails from screen readers

mtribone:
*** THERE MAY BE A BETTER WAY ***

Override the featured and recent views on the home page to add aria hidden true to the thumbnail image options. Updates the catalog views from Blacklight to hide the thumb via aria hidden. Updates the works and collections listings in the MY index partials.

Adds empty alt tag as an option for render thumbnail tag. Update spec test to switch i element to span element per Bootstrap styles.

### DIFF
--- a/app/views/catalog/_index_gallery.html.erb
+++ b/app/views/catalog/_index_gallery.html.erb
@@ -1,0 +1,8 @@
+<div class="document col-xs-6 col-md-3">
+  <div class="thumbnail">
+    <%= render_thumbnail_tag(document, { alt: '', "aria-hidden" => true }, :counter => document_counter_with_offset(document_counter)) %>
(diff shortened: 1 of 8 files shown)


carolyncole (changes requested):
@mtribone I wonder if overriding render_thumbnail_tag helper would be the easier way to hid all thumbnails from screen readers.

mtribone:
Yeah @awead and Derek and I worked on that on Friday in our refactoring session. Just trying to get the tests to run/pass. My branch is here:

https://github.com/psu-stewardship/scholarsphere/tree/thumbnails-%23464-3

mtribone:
Got a better way to do this. Abandon ship!

PR #536, opened 2017-01-11
Override uploader.js

### LINKED ISSUE #453: File Uploaded many small files fails
In ScholarSphere 3.0
When I create a new work
And I choose many (10 or more) small (15KB or less) files using add files or add folder
Then I get some blank lines where a file should be in the upload screen

awead:
Requires some rearranging of the javascript so we can limit concurrent uploads to only 1 file at a time.

This fixes #453 but it's not ideal. The ideal solution would be to have multiple concurrent uploads in our current deployment, however, I'm not sure if that possible give server setup, etc.

If you all are okay with this, we can merge. You can try it out on QA. It doesn't slow things down too much to do one file at a time. Let me know what you think: @cam156 @mtribone @olendorf

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -8,5 +8,10 @@
 // Required by Blacklight
 //= require blacklight/blacklight
 
-//= require_tree .
+// Requiring each file individually so we can load Sufia's JS, with our overrides, at the end.
+//= require batch_edit
+//= require blacklight_gallery
+//= require openseadragon
+//= require scholarsphere_fileupload
+//= require scholarsphere_layout
 //= require sufia
--- a/app/assets/javascripts/sufia/uploader.js
+++ b/app/assets/javascripts/sufia/uploader.js
@@ -1,0 +1,74 @@
+// This is overriding Sufia's uploader.js so we can change the value of limitConcurrentUploads
+
+//= require fileupload/tmpl
+//= require fileupload/jquery.iframe-transport
+//= require fileupload/jquery.fileupload.js
+//= require fileupload/jquery.fileupload-process.js
+//= require fileupload/jquery.fileupload-validate.js
+//= require fileupload/jquery.fileupload-ui.js
+//
+/*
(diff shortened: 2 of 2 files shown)


app/assets/javascripts/sufia/uploader.js
```diff
@@ -0,0 +1,71 @@
...
+
+  $.fn.extend({
+    sufiaUploader: function( options ) {
+      // Initialize our jQuery File Upload widget.
+      // TODO: get these values from configuration.
+      this.fileupload($.extend({
+        // xhrFields: {withCredentials: true},              // to send cross-domain cookies
+        // acceptFileTypes: /(\.|\/)(png|mov|jpe?g|pdf)$/i, // not a strong check, just a regex on the filename
+        // limitMultiFileUploadSize: 500000000, // bytes
+        limitConcurrentUploads: 1,
```
carolyncole, line 33: @awead is this the line that makes all the difference?

app/assets/javascripts/application.js
```diff
@@ -8,5 +8,9 @@
 // Required by Blacklight
 //= require blacklight/blacklight
 
-//= require_tree .
-//= require sufia
\ No newline at end of file
+//= require batch_edit
```
carolyncole, line 12: @awead is this removal of require_tree needed for the PR or just a preference?

carolyncole (changes requested):
@awead The changes look ok.  Some comments either at the top or around the 1 to let us know what we changed and why would be nice.  We do not want to do a comparison a month or two from now and think we do not need this override.

app/assets/javascripts/application.js
awead: @cam156 require_tree was causing all kinds of JS loading errors.

app/assets/javascripts/sufia/uploader.js
awead: yep. I'll add a comment.

Issue #526, opened 2017-01-04
Edit Work <head><title> Content can be Better


mtribone:
"Web pages should include a TITLE tag within the HEAD area, and the title should be unique to each page on the Web site. The TITLE can be repeated as the H1 of the page."

http://accessibility.psu.edu/titleshtml/

We could do a little better job with the `<head><title>` and the work `<h1>` even though they both can be the same. Per the example on the accessibility site, we could give the `<head><title>` some context about its relationship in the site.

The `<head><title>` when editing a work shows both the work title and its work identifier.
"Star Wars Interfaces // Work [70r967436w] // ScholarSphere"

Perhaps a better example:
"Edit Work | Star Wars Interfaces | ScholarSphere"

carolyncole:
Changed the work head title to do the following:
1. Use the bar separator instead of the double //
1. First node should include page action (show vs edit)
1. Second Node should be title
1. Third node should be "Work id: "id""
1. Forth Node should be scholarSphere

`Edit Work| Star Wars Interfaces| Word id: 70r967436w| ScholarSphere`
`Work| Star Wars Interfaces| Word id: 70r967436w| ScholarSphere`

Make the collection title constant with work.

Issue #515, opened 2016-12-20
Work Show Page Needs Better <head><title> and Page H1


mtribone:
"Web pages should include a TITLE tag within the HEAD area, and the title should be unique to each page on the Web site. The TITLE can be repeated as the H1 of the page."

http://accessibility.psu.edu/titleshtml/

We could do a little better job with the `<head><title>` and the work `<h1>` even though they both can be the same. Per the example on the accessibility site, we could give the `<head><title>` some context about its relationship in the site.

Examples
1.  "I am the Work Title | ScholarSphere"
1.  "I am the Work Title | Works | ScholarSphere"

PR #538, opened 2017-01-11
Update <head><title> in Works and Collections for Better Accessibility

mtribone:
Fixes https://github.com/psu-stewardship/scholarsphere/issues/526 and https://github.com/psu-stewardship/scholarsphere/issues/515

Overrides construct_page_title method  and replaces `//` with ` | ` and defines page_title in work show presenter.

Adds the collection show page so that we can override the page title to keep it consistent with the style of the work show page.

Updates edit page for collections to follow style from work edit page.

Removes the construct_page_title method to just pass a string with dynamic elements to output contents of title element.

### DIFF
--- a/app/helpers/title_helper.rb
+++ b/app/helpers/title_helper.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+module TitleHelper
+  include CurationConcerns::TitleHelper
+
+  def construct_page_title(*elements)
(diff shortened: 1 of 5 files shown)


mtribone:
In the views, I opted to skip the construct_page_title method and just stick with text and interpolated values. Holla back yo!

Issue #461, opened 2016-11-17
Strange errors in the dashboard with proxy deposits


carolyncole:
New Note 4

2016-11-17 11:53:53 -0500 (30460) 
ActiveFedora::ActiveFedoraError (Model mismatch. Expected CurationConcerns::WorkRelation::DummyModel. Got: FileSet):                                            
  active-fedora (10.3.0.rc2) lib/active_fedora/relation/finder_methods.rb:204:in `class_to_load'                                                                
  active-fedora (10.3.0.rc2) lib/active_fedora/relation/finder_methods.rb:195:in `load_from_fedora'                                                             
  active-fedora (10.3.0.rc2) lib/active_fedora/relation/finder_methods.rb:181:in `find_one'                                                                     
  active-fedora (10.3.0.rc2) lib/active_fedora/relation/finder_methods.rb:228:in `find_with_ids'                                                                
  active-fedora (10.3.0.rc2) lib/active_fedora/relation/finder_methods.rb:52:in `find'                                                                          
  active-fedora (10.3.0.rc2) lib/active_fedora/relation/finder_methods.rb:85:in `exists?'                                                                       
  /opt/heracles/deploy/scholarsphere/shared/bundle/ruby/2.3.0/bundler/gems/sufia-35c01618b4d3/app/models/proxy_deposit_request.rb:83:in `deleted_work?'

carolyncole:
It seems that if you have old proxies where the item no longer exists you get a very strange error about model miss match.  Once I deleted my proxy deposit request all was well.  That said we should look at how those request are being migrated and make sure they work.

carolyncole:
Handle any StandardException (Runtime or Other) instead of erroring

awead:
Steps to recreate:

1. bundle exec rake dev:clean db:reset
2. bundle exec rails s
3. log in and create a work
4. Visit the file set page for the work and copy the file set id from the url
5. bundle exec rails c
6. user = User.create
7. ProxyDepositRequest.create(work_id: [fileset id], sending_user_id: [your user id], receiving_user_id: user.id)
8. Visit http://localhost:3000/dashboard with your user id logged in

carolyncole:
@awead was there a change to handle the error?  Or did we decide not to handle the error?  What makes this ticket closed?

awead:
@cam156 I've rescued the error here:

[...]

PR #549, opened 2017-01-16
DO NOT MERGE: Override ProxyDepositRequest

awead:
This is to catch ActiveFedoraError model mismatch when a work is expected but we get a fileset instead. Ref #461 

@cam156 @mtribone This fixes the issue we've been seeing with the strange errors, but in order to do this, we have to override the entire ProxyDepositRequest model. It isn't namespaced like the controllers and other code so we can't use inheritance.

I believe the reason this was happening in the first place was that there were leftover records in the table that were once works but were reused as file sets. So, I'm not really sure what the likelihood is of this occurring during real use. Re-writing an entire class just to add 2 lines for this seems like overkill.

What do you all think?

### DIFF
--- a/app/models/proxy_deposit_request.rb
+++ b/app/models/proxy_deposit_request.rb
@@ -1,0 +1,114 @@
+# frozen_string_literal: true
+class ProxyDepositRequest < ActiveRecord::Base
+  include Blacklight::SearchHelper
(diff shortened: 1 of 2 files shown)


awead:
Do not merge until we've done a full import!

PR #555, opened 2017-01-17
File Set <head><title> for Edit and Show

### LINKED ISSUE #516: File Set Show <head><title> Content Can be Better
"Web pages should include a TITLE tag within the HEAD area, and the title should be unique to each page on the Web site. The TITLE can be repeated as the H1 of the page."

http://accessibility.psu.edu/titleshtml/

We could do a little better job with the `<head><title>` and the work `<h1>` even though they both can be the same. Per the example on the accessibility site, we could give the `<head><title>` some context about its relationship in the site.

Currently, the file set `<head><title>` doesn't provide any meaningful context. It is literally `Show File Set // ScholarSphere`.

Better Example
1.  "File Set Title | I am the Work Title | ScholarSphere"

mtribone:
Adds page_title method to file set presenter, provides page_title on show, and adds same to edit page. Fixes #516

### DIFF
--- a/app/views/curation_concerns/file_sets/edit.html.erb
+++ b/app/views/curation_concerns/file_sets/edit.html.erb
@@ -1,0 +1,35 @@
+<%
+  # Overrides Sufia app > views > curation_concerns > file_sets > edit.html.erb
+  # provide :page_title, curation_concern_page_title(curation_concern)
+%>
+<% provide :page_title, "Edit File | #{curation_concern} | File ID: #{curation_concern.to_param} | ScholarSphere" %>
+<% provide :page_header do %>
+    <h1 class="lower">Edit <%= curation_concern %></h1>
+<% end %>
+
+<div class="row">
+  <div class="col-xs-12 col-sm-4">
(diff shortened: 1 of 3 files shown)


app/views/curation_concerns/file_sets/edit.html.erb
```diff
@@ -0,0 +1,31 @@
+<% provide :page_title, "Edit File | #{curation_concern} | File ID: #{curation_concern.to_param} | ScholarSphere" %>
```
carolyncole, line 5: @mtribone I'm assuming this is an override of CC or sufia. Can you add a comment what was changed so that when we want to compare later it will be easier?

app/views/curation_concerns/file_sets/show.html.erb
```diff
@@ -0,0 +1,21 @@
+<% provide :page_title, @presenter.page_title %>
```
carolyncole, line 2: Simmilar request ehere

carolyncole (changes requested):

Issue #513, opened 2016-12-20
Alternative Text for the Penn State Mark


mtribone:
Typically, alternative text for images should describe the contents and/or the function of the image. However, since this is a mark or logo inside of a link and in a consistent spot a slight change to the text is need. The text "Penn State mark" can be changed to "Penn State" for better support.

http://webaim.org/techniques/alttext/#logos

PR #556, opened 2017-01-18
About, Help <head><title> and Penn State Mark

### LINKED ISSUE #512: About and Help Pages Need Better <head><title> and Page H1
"Web pages should include a TITLE tag within the HEAD area, and the title should be unique to each page on the Web site. The TITLE can be repeated as the H1 of the page."

http://accessibility.psu.edu/titleshtml/

About Page
`<title>Show Page // ScholarSphere</title>`
`<h1>` is missing

Help Page
`<title>Help Static // ScholarSphere</title>`
`<h1>Help</h1>`

mtribone:
Provides the page title for the about and help pages. Updates URL and alt text for Penn State mark. Fixes #512 and #513

### DIFF
--- a/app/controllers/static_controller.rb
+++ b/app/controllers/static_controller.rb
@@ -2,6 +2,10 @@
 class StaticController < ApplicationController
   rescue_from AbstractController::ActionNotFound, with: :render_404
 
+  def about
+    @page = ContentBlock.find_or_create_by(name: "about")
+  end
+
   def help
     @page = ContentBlock.find_or_create_by(name: "help_faq")
   end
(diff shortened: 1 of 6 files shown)


app/views/pages/show.html.erb
```diff
@@ -0,0 +1,3 @@
+<% provide :page_title, "About | ScholarSphere" %>
```
carolyncole: @mtribone I wonder if this page is used for more than just the about page.  It looks to be a generic show page for content block.  I wonder if the @page variable has more info

mtribone: It _seems_ that this is the only use for it, but it doesn't mean that it couldn't be used for other things in the future. I'll investigate.

PR #557, opened 2017-01-18
Improve accessibility for social media sharing

mtribone:
Adds an additional sr-only span because the title attribute for hyperlinks are not read consistently by all assistive technologies.


carolyncole:
@mtribone can you update the comment at the top to indicate that the file can not be removed for just the PR?

PR #617, opened 2017-02-07
Toolbar displays when not logged in on small screens

### LINKED ISSUE #484: Toolbar displays on mobile or small screens when not logged in
When I am not logged into ScholarSphere and am viewing the site on a mobile device or small screen, the toolbar with the hamburger menu displays without menu items. I would expect that the toolbar be hidden from view just like the behavior on larger screens.

screen shot 2016-12-06 at 9 29 07 am

mtribone:
If a user is signed in, then they get the collapsable toolbar. Otherwise don't render the toolbar. Fixes #484

### DIFF
--- a/app/views/_masthead.html.erb
+++ b/app/views/_masthead.html.erb
@@ -12,6 +12,7 @@
         </div>
       </div>
     </div>
+    <% if user_signed_in? %>
     <div class="row ss-toolbar">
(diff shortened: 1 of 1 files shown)


mtribone:
screen shot 2017-02-07 at 2 14 59 pm

Issue #554, opened 2017-01-17
Merge scholarsphere.yml config into application.yml


awead:
This would refactor our current config method that uses `config/scholarsphere.yml` and replace it with `config/application.yml` which Figaro uses. The Figaro gem was installed to manage the temp. directory environment variable, but could be used to manage any other kind of system-specific variable.

awead:
See related: #542 #553

carolyncole:
Change application.rb to use figaro and move the information into application.yml instead of scholarsphere.yml

PR #648, opened 2017-02-13
Move scholarsphere.yml to application.yml, ref #554

awead:


config/application.rb
```diff
@@ -93,6 +92,6 @@ class Application < Rails::Application
     config.active_record.raise_in_transactional_callbacks = true
 
     config.action_mailer.default_options = { from: "umg-up.its.sas.scholarsphere-email@groups.ucs.psu.edu" }
-    config.action_mailer.default_url_options = { host: ss_config.fetch(:service_instance, Socket.gethostname), protocol: 'https' }
+    config.action_mailer.default_url_options = { host: ENV.fetch("service_instance", Socket.gethostname), protocol: 'https' }
```
carolyncole: At this line should you just use 'config.service_instance'

config/application.rb
```diff
@@ -76,7 +75,7 @@ class Application < Rails::Application
     # Configure sensitive parameters which will be filtered from the log file.
     config.filter_parameters += [:password]
 
-    config.stats_email = ss_config.fetch(:stats_email, "ScholarSphere Stats <umg-up.its.sas.scholarsphere-email@groups.ucs.psu.edu>")
+    config.stats_email = ENV.fetch("stats_email", "ScholarSphere Stats <umg-up.its.sas.scholarsphere-email@groups.ucs.psu.edu>")
```
carolyncole: At this line should you use config.stats_email?  It is set above

carolyncole (changes requested):
@awead two small changes to consolidate where we are accessing the envs.

config/application.rb
awead: @cam156 I don't understand. Above where? do you mean stats_from_email?

carolyncole: @awead I assumed that since there were five tests and five variables in the sample and five assigns above that all five variables were getting set in the original block.  It looks like derivatives path is untested and I would like for you to move the assign up to the main block for consistency.

config/application.rb
```diff
@@ -23,13 +23,12 @@ class Application < Rails::Application
...
-    ss_config = YAML.load(File.read(File.join(Rails.root, 'config/scholarsphere.yml')))[Rails.env].with_indifferent_access
-
-    config.ffmpeg_path = ss_config.fetch(:ffmpeg_path, "ffmpeg")
-    config.derivatives_path = ss_config.fetch(:derivatives_path, File.join(Rails.root, 'tmp', 'derivatives'))
-    config.service_instance = ss_config.fetch(:service_instance, Socket.gethostname)
-    config.virtual_host = ss_config.fetch(:virtual_host, "https://#{Socket.gethostname}")
-    config.google_analytics_id = ss_config.fetch(:google_analytics_id, nil)
+    # Environment variables list here are defined in application.yml
+    config.ffmpeg_path = ENV.fetch("ffmpeg_path", "ffmpeg")
+    config.derivatives_path = ENV.fetch("derivatives_path", File.join(Rails.root, 'tmp', 'derivatives'))
```
carolyncole, line 28: This variable is not in the sample files and is untested.

config/application.rb
awead: @cam156 derivatives_path is tested, it's just using the default specified in `fetch`. I've added a key to the yaml file though.

config/application.rb
awead: @cam156 it's tested spec/config/application_spec.rb

carolyncole (approved):

carolyncole:
@awead I just went ahead and ticket the issues that I had with this PR.  That way we will not loose the fact that things are a bit off, but we can also get the code changes into the system.

Issue #645, opened 2017-02-13
Change Date Created to Published Date


carolyncole:
Changed the label Date Created to Published Date across the system

PR #649, opened 2017-02-13
Change Date Created to Published Date, ref #645

awead:


spec/factories/works.rb
```diff
@@ -41,18 +41,6 @@
       end
     end
 
-    trait :full_meta_data do
```
carolyncole, line 44: @awead not sure why you are removing this trait.  That seems a bit weird.

carolyncole (changes requested):
@awead looks good with the exception of removing the trait.

spec/factories/works.rb
awead: @cam156 it's not used.

carolyncole: @awead it is not really related to this commit, but it does look like it should have been removed in this commit or an earlier one( https://github.com/psu-stewardship/scholarsphere/commit/ca52ec1f5a74441168ef17ac61bd6a61da3381e5#diff-8011362fff9e5db5c9997382f4418f28  ) Just making sure we are not loosing test coverage.

carolyncole (approved):

Issue #632, opened 2017-02-09
Facets should be case insensitive


olendorf:
Having case sensitive sorting of facets may not be the best way to do it.

Also how to handle a clever reordering of names yielding meaning

carolyncole:
Change the ALL 9 solr facet result to be white space, punctuation and case agnostic.
"Patricia  Hswe" ~= "Patricia Hswe"
"PATRICIA HSWE" ~= "Patricia Hswe"

This may be accomplished in the solr schema or via another solr field created by to_solr.  We do not have a preference, but if solr changes then the chef scripts will need to be updated.

PR #655, opened 2017-02-15
Cleanup creator and other facets prior to indexing

awead:
This is a solution to #632. I'm not too keen on it.

The initial solution is simple: cleanup facet terms before sending them to solr such that _ADAM WEAD_ becomes _Adam Wead_. However, it gets complicated because we have linked facets, urls that map to catalog searches with a given faceted search. Because we've changed the facet but not the actual value the user entered, the links don't work.

The solution then is to modify the links so that they display the original user's entered text, but link to the modified facet term.

This is kind of icky because we're essentially modifying twice: first on the way in (to Solr) and then on the way out again when we display. The solution below could be made more efficient, perhaps. But it strikes me that the _ideal_ solution might be to modify the user input in the first place, normalizing values upfront.

Thoughts? @cam156 @olendorf ?

### DIFF
--- a/app/services/field_configurator.rb
+++ b/app/services/field_configurator.rb
@@ -3,12 +3,12 @@ class FieldConfigurator
   def self.common_fields
     {
       resource_type: FieldConfig.new("Resource Type"),
-      creator: FieldConfig.new("Creator"),
-      keyword:  FieldConfig.new("Keyword"),
+      creator: FieldConfig.new(label: "Creator", facet_cleaners: [:titleize]),
+      keyword:  FieldConfig.new(label: "Keyword", facet_cleaners: [:downcase]),
       subject: FieldConfig.new("Subject"),
       language: FieldConfig.new("Language"),
       based_near: FieldConfig.new("Location"),
-      publisher: FieldConfig.new("Publisher"),
+      publisher: FieldConfig.new(label: "Publisher", facet_cleaners: [:titleize]),
       file_format: FieldConfig.new("File Format")
     }
   end
--- a/app/services/solr_document_groomer.rb
+++ b/app/services/solr_document_groomer.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+# Cleans up existing fields in the solr document before it is sent on for indexing
+class SolrDocumentGroomer
+  attr_reader :document
+
+  # @param [SolrDocument, Hash]
+  # @return [SolrDocument, Hash]
+  def self.call(document)
+    new(document).groom
+  end
+
+  def initialize(document)
+    @document = document
+  end
+
+  def groom
+    FieldConfigurator.facet_fields.each do |field, config|
+      cleaned_fields = FacetValueCleaningService.call(document.fetch(Solrizer.solr_name(field.to_s, :facetable), []), config)
+      document[Solrizer.solr_name(field.to_s, :facetable)] = cleaned_fields
+    end
+  end
+end
--- a/app/views/curation_concerns/base/_attribute_rows.html.erb
+++ b/app/views/curation_concerns/base/_attribute_rows.html.erb
@@ -1,11 +1,19 @@
-<%= presenter.attribute_to_html(:creator, render_as: :faceted) %>
+<%= presenter.attribute_to_html(:creator, render_as: :translated_facet,
+                                          mapping: presenter.facet_mapping(:creator)) %>
+
 <%= presenter.attribute_to_html(:contributor, render_as: :faceted) %>
 <%= presenter.attribute_to_html(:admin_set, render_as: :faceted) %>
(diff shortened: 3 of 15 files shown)


app/services/solr_document_groomer.rb
```diff
@@ -0,0 +1,20 @@
...
+    new(document).groom
+  end
+
+  def initialize(document)
+    @document = document
+  end
+
+  def groom
+    clean_names = FacetValueCleaningService.call(document.fetch(Solrizer.solr_name('creator', :facetable), []))
+    document[Solrizer.solr_name('creator', :facetable)] = clean_names
```
carolyncole: @awead I was thinking our changes would be Additive, not changing.  
```
document[Solrizer.solr_name('creator_cleaner', :facetable)] = clean_names
```
We would then just use that for our facets and leave the original documents alone.

awead:
ping @cam156 This is ready for review now. Note, I've added another commit for #659. I will also be doing a separate PR and ticket to address the cleanup of spaces in fields because it's really separate from the scope of this ticket. Let me know what you think.

app/services/field_configurator.rb
```diff
@@ -3,20 +3,20 @@ class FieldConfigurator
   def self.common_fields
     {
       resource_type: FieldConfig.new("Resource Type"),
-      creator: FieldConfig.new("Creator"),
-      keyword:  FieldConfig.new("Keyword"),
+      creator: FieldConfig.new(label: "Creator", facet_values: [:titleize]),
```
carolyncole: @awead 'facet_values' name really threw me.  facet_cleaner_methods or facet_value_cleaners might be a better name.   When I saw the name facet_values in the code above I could not understand why you were calling send on a value of the facet.

app/services/field_configurator.rb
```diff
@@ -3,20 +3,20 @@ class FieldConfigurator
...
-      file_format: FieldConfig.new("File Format")
+      publisher: FieldConfig.new(label: "Publisher", facet_values: [:titleize]),
+      file_format: FieldConfig.new("File Format"),
+      contributor: FieldConfig.new(label: "Contributor", facet_values: [:titleize])
     }
   end
 
   def self.index_fields
     { description: FieldConfig.new("Description") }.merge(
-      common_fields).merge(contributor: FieldConfig.new("Contributor"),
```
carolyncole: @awead why did contributor get removed from the index?  Was this planned?

app/views/curation_concerns/base/_attribute_rows.html.erb
```diff
@@ -1,11 +1,11 @@
-<%= presenter.attribute_to_html(:creator, render_as: :faceted) %>
-<%= presenter.attribute_to_html(:contributor, render_as: :faceted) %>
+<%= presenter.attribute_to_html(:creator, render_as: :valued_facet, facet_values: presenter.facet_values(:creator)) %>
+<%= presenter.attribute_to_html(:contributor, render_as: :valued_facet, facet_values: presenter.facet_values(:contributor)) %>
 <%= presenter.attribute_to_html(:admin_set, render_as: :faceted) %>
 <%= presenter.attribute_to_html(:subject, render_as: :faceted) %>
-<%= presenter.attribute_to_html(:publisher, render_as: :faceted) %>
+<%= presenter.attribute_to_html(:publisher, render_as: :valued_facet, facet_values: presenter.facet_values(:publisher)) %>
```
carolyncole: @awead here it seems like the facet_values is the actual value.  Is that correct?

spec/presenters/work_show_presenter_spec.rb
```diff
@@ -55,4 +55,10 @@
       it { is_expected.not_to be_uploading }
     end
   end
+
+  describe "#facet_values" do
+    let(:work) { build(:work, creator: ["JOE SMITH"]) }
+    subject { presenter.facet_values(:creator) }
+    it { is_expected.to eq("JOE SMITH" => "Joe Smith") }
```
carolyncole: Here is a good example of my confusion. This line is actual values, and `https://github.com/psu-stewardship/scholarsphere/pull/655/files#diff-ee51de3cf9faf45a5042f3c4059047f1R59` for the same name is cleaning methods.

carolyncole (changes requested):

app/services/field_configurator.rb
awead: @cam156 it didn't. It's been moved to common_fields so it can be both an index and facet field. See #659

spec/presenters/work_show_presenter_spec.rb
awead: @cam156 good point. I'll change the config key to be something like `:facet_value_cleaners` and leave the presenter method as `facet_values` because it is returning actual values there.

carolyncole (approved):

Issue #644, opened 2017-02-10
Update application.yml


awead:
Update the application.yml file on the  QA servers. See http://sites.psu.edu/dltdocs/?p=3521. Ref #553

carolyncole:
#553 includes the documentation previously created.  The changes occurred during #554.
Change the application.yml to include the information in the scholarsphere.yml becuase that information is now being read from the application.yml.

PR #656, opened 2017-02-16
Validate all our configuration files prior to deploying, ref #644

awead:
@cam156 here's a suggestion avoid future problems with our configuration files on the servers. This will check for the presence of keys in our `application.yml` on the server during the Capistrano deploy process and if the keys don't matchup, the deploy aborts.

Currently, it's only checking the one yaml file, but it could be expanded to check others. The required keys are defined in `lib/scholarsphere/config.rb` which might not be the best place.

### DIFF
--- a/config/application.yml
+++ b/config/application.yml
@@ -1,10 +1,9 @@
 # Figaro configuration file
 # For further instructions see: http://sites.psu.edu/dltdocs/?p=3521
+# Each key under production is checked in Scholarsphere::Config to ensure that files on servers
+# are correct. Add or removing any keys here must include a corresponding change to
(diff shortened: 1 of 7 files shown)


carolyncole:
@awead did you already deploy this to QA to check if it works?

awead:
@cam156 not yet. I wasn't sure if it was a good time yesterday. I can test it out Monday.

awead:
@cam156 I deployed to qa and tested it. If you add or remove any variable keys from the config file on the server, then the cap deploy will fail. We can expand the process to include other config files later, if we want, but for now, if there's anything missing from our application.yml file or if it's different in any way, cap will let us know.

PR #662, opened 2017-02-22
return batches with only one item since there are no relations to be made

carolyncole:
Additionally the save is not needed, the tests pas with or without it.

This is an effort to make the batch processing faster.  It may or may not make larger batches faster.

### DIFF
--- a/lib/scholarsphere/import/batch_translator.rb
+++ b/lib/scholarsphere/import/batch_translator.rb
@@ -6,10 +6,10 @@ module Import
 
       def build_from_json(json)
         work_ids = json['generic_file_ids']
+        return if work_ids.length < 2 # skip anything that doesn't have at least two works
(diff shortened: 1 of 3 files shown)


lib/scholarsphere/import/batch_translator.rb
```diff
@@ -6,10 +6,10 @@ class BatchTranslator < Sufia::Import::Translator
 
       def build_from_json(json)
         work_ids = json['generic_file_ids']
+        return if work_ids.length <= 1 # no relations unless we have atleasrt two works
```
awead: how about `# skip anything that doesn't have at least two works`

awead (changes requested):

carolyncole:
@awead updated the comment and the code to reflect your suggested change

awead (approved):

PR #669, opened 2017-02-23
Display message after clicking Save

awead:
Friday. Begins with F, and that stands for **Fancy**!

Here's a video of the interaction with the screen reader voice over. You'll need to turn your volume up to 11 to hear it.

https://www.dropbox.com/s/u2858kwnplbv9a0/demo.mov?dl=0

Might need some css tweaks. @mtribone @cam156


app/assets/javascripts/scholarsphere_fileupload.js
```diff
@@ -1,15 +0,0 @@
-//500 MB  max file size 
-max_file_size = 500000000;
```
carolyncole, line 2: @awead why did you remove the maximums?  These override items in curation concerns https://github.com/projecthydra/curation_concerns/blob/master/app/assets/javascripts/curation_concerns/uploader.js#L13

app/views/curation_concerns/base/_form_progress.html.erb
```diff
@@ -0,0 +1,56 @@
+<aside id="form-progress" class="form-progress panel panel-default">
```
carolyncole, line 1: Can you add a comment to the top of the class to describe why we are overriding?

carolyncole (changes requested):
@awead looks good to me.  I like that it speaks!  Just a couple little tweaks.

app/assets/javascripts/scholarsphere_fileupload.js
awead: @cam156 these don't do anything anymore. The size restriction is now in:
https://github.com/psu-stewardship/scholarsphere/blob/develop/app/assets/javascripts/sufia/uploader.js#L34-L35

app/views/curation_concerns/base/_form_progress.html.erb
```diff
@@ -0,0 +1,56 @@
...
+    <br>
+    <%= link_to t(:'helpers.action.cancel'),
+                sufia.dashboard_index_path,
+                class: 'btn btn-default' %>
+    <%= f.submit 'Save', class: 'btn btn-primary', onclick: "confirmation_needed = false;", id: "with_files_submit", name: "save_with_files" %>
+
+  </div>
+  <div class="panel-footer hidden">
+    <div class="progress">
+      <div class="progress-bar progress-bar-striped active" style="width: 100%">
```
mtribone: Can the inline style for the width be applied to a class that you are using?

app/views/curation_concerns/base/_form_progress.html.erb
awead: @cam156 I put the comment above the spot at which we're overriding.

app/views/curation_concerns/base/_form_progress.html.erb
awead: @mtribone done.

carolyncole (approved):

PR #680, opened 2017-02-28
Update Relationships and Descriptions on Work Show to use Description List

mtribone:
Grabs the relationships partials so that we can start to override the tables with lists for relationships and descriptions.

Overrides the attribute renderer and metadata partial and swaps out table for definition lists. Fixes spec tests for by swapping out table-related code with dl.

Add Bootstrap row and columns for relationships and descriptions, adds CSS to add comma and space between inline metadata.

Swaps the columns of relationships and descriptions, adds classes for dt and dd to align them inline.

### DIFF
--- a/app/renderers/curation_concerns/renderers/attribute_renderer.rb
+++ b/app/renderers/curation_concerns/renderers/attribute_renderer.rb
@@ -1,0 +1,71 @@
+# frozen_string_literal: false
+# Overrides default renderer to replace tables with description lists for accessibility
+require "rails_autolink/helpers"
+
+module CurationConcerns
+  module Renderers
+    class AttributeRenderer
+      include ActionView::Helpers::UrlHelper
+      include ActionView::Helpers::TranslationHelper
+      include ActionView::Helpers::TextHelper
+      include ConfiguredMicrodata
+
+      attr_reader :field, :values, :options
+
+      # @param [Symbol] field
+      # @param [Array] values
+      # @param [Hash] options
+      def initialize(field, values, options = {})
+        @field = field
+        @values = values
+        @options = options
+      end
+
+      # Create definition terms and descriptions for the attribute
+      def render
+        markup = ''
+
+        return markup if !values.present? && !options[:include_empty]
+        markup << %(<dt class="attribute-term">#{label}</dt>)
+        attributes = microdata_object_attributes(field).merge(class: "attribute #{field}")
+        Array(values).each do |value|
+          markup << "<dd#{html_attributes(attributes)}>#{attribute_value_to_html(value.to_s)}</dd>"
+        end
+        # markup << %(</dd>)
+        markup.html_safe
+      end
+
+      # @return The human-readable label for this field.
+      # @note This is a central location for determining the label of a field
+      #   name. Can be overridden if more complicated logic is needed.
+      def label
+        translate(
+          :"blacklight.search.fields.show.#{field}",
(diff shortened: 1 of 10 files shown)


mtribone:
screen shot 2017-02-28 at 10 02 33 am

app/renderers/curation_concerns/renderers/attribute_renderer.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: false
```
carolyncole, line 1: Can you add a comment at the top to explain why you overrode the Attribute renderer?

Something like ` switching default renderer to show list item instead of table row`

app/views/curation_concerns/base/_metadata.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="col-xs-12 col-sm-7">
```
carolyncole, line 2: Can you add a similar comment to the top of this file?

app/views/curation_concerns/base/_relationships.html.erb
```diff
@@ -0,0 +1,4 @@
+<div class="col-xs-12 col-sm-5">
```
carolyncole, line 1: Is this override necessary?  I do not see any table or list references here.

app/views/curation_concerns/base/_relationships_parent_row.html.erb
```diff
@@ -0,0 +1,12 @@
+ <h3><%= t(".label", type: type.humanize) %></h3>
```
carolyncole: Add a comment about changing table to list here too.

app/views/curation_concerns/base/_relationships_parent_row_empty.html.erb
```diff
@@ -0,0 +1,2 @@
+<h3><%= t(".label", type: type.humanize) %></h3>
```
carolyncole, line 2: Is this override needed

app/views/curation_concerns/base/show.html.erb
```diff
@@ -0,0 +1,31 @@
+<% provide :page_title, @presenter.page_title %>
```
carolyncole, line 2: I assuming this override was required for your css changes?

carolyncole (changes requested):
@mtribone looks good.  Just need a little documentation so we do not lose your changes in the future.  Since the overrides are very similar to the originals we might just think they were added for no reason and delete them later without comments.

app/views/curation_concerns/base/_relationships.html.erb
mtribone: This override is for the Bootstrap layout changes for xs and sm

app/views/curation_concerns/base/_relationships_parent_row_empty.html.erb
mtribone: yup. It also removes some table cruft. https://github.com/projecthydra/sufia/blob/7.2-migration/app/views/curation_concerns/base/_relationships_parent_row_empty.html.erb

app/views/curation_concerns/base/show.html.erb
mtribone: Swaps the order of relationships and descriptions for better layout

carolyncole (approved):

PR #682, opened 2017-03-01
Page Titles for Works and File Sets and File Set Titles in Work Show

mtribone:
Removes the page_title method from the file set presenter and updates the show page for the file set. Removes the page_title method from the work show presenter and adds the page title to the CC show page.


mtribone:
screen shot 2017-03-01 at 1 58 33 pm

carolyncole (approved):

Issue #601, opened 2017-02-02
Add Help for Large Files should be above all files


carolyncole:
Move the green help area above the all the file upload on the Files tab

carolyncole:
<img width="965" alt="screen shot 2017-02-27 at 9 15 40 am" src="https://cloud.githubusercontent.com/assets/1599081/23364513/5f9d3034-fccd-11e6-9d91-c86aaa19a2e2.png">

PR #683, opened 2017-03-01
Moving the help to the top of the page and changing the cloud wording

carolyncole:
refs #601
screen shot 2017-03-01 at 9 51 31 am

### DIFF
--- a/app/assets/stylesheets/scholarsphere/add_edit_work.scss
+++ b/app/assets/stylesheets/scholarsphere/add_edit_work.scss
@@ -14,11 +14,6 @@ visually have a focus just like the other form elements.
   overflow: hidden;
 }
 
-.inputfile + label:before {
-  font-family: FontAwesome;
-  content: "\f067"
-}
-
 .inputfile + label {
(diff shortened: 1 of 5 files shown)


carolyncole:
@mtribone @olendorf what do you think?

mtribone:
@cam156 I think that the header for cloud files is much more readable now. I wonder if people will actually read the help text, but that is for users to tell us. Part of me wonders if the button should also be "+ Add files" for consistency. Add versus browse seems more active to me, but not a blocker.

carolyncole:
@mtribone how about "+ Add Cloud Files" to make the buttons read differently?

carolyncole:
@mtribone adding the + is particularly tricky for the browse everything button.  Would just removing it from all work?

screen shot 2017-03-03 at 10 57 34 am

mtribone:
The more that we can remove items that do not add or increase the meaning of something in the UI, the better the interface will be for users.

carolyncole:
@awead & @mtribone this should be all set now.

Issue #508, opened 2016-12-19
CreateDerivativesJob is creating temp file with spaces


carolyncole:
CreateDerivativesJob is creating a temp file with space which is failing in resque

```
Worker
ssjobs1qa:16823 on files at 2016/12/19 09:37:34 -0500
Retry or Remove
Class
ActiveJob::QueueAdapters::ResqueAdapter::JobWrapper
Arguments
---
job_class: CreateDerivativesJob
job_id: 57e1e60a-04b1-406f-a93e-a2716e40672d
queue_name: files
arguments:
- _aj_globalid: gid://scholar-sphere/FileSet/gf06g2885
- gf06g2885/files/77789ea1-9e7f-4c2a-aa4e-35375e4c9bfd
- "/tmp/gf06g2885_version1"
locale: en
Exception
URI::InvalidURIError
Error
bad URI(is not URI?): /tmp/2013-11-19 Memo for PPE1 interview.pdf
```

carolyncole:
Also we have ones with commas & dashes:
```
bad URI(is not URI?): /tmp/CV, Nelson.pdf
...
bad URI(is not URI?): /tmp/SP2-Jennifer Pressley Revised.pdf
```

carolyncole:
What seems to be happening here is that the file is expected to exists, but it does not.  This is causing the bad URI and  `No such file or directory @ rb_sysopen `

carolyncole:
This error is no longer occurring during the import

carolyncole:
Unfortunately we got one in the last import:

```
---
job_class: ImportVersionJob
job_id: c0d30855-53c4-4259-9413-6456788a8e13
queue_name: files
arguments:
- _aj_globalid: gid://scholar-sphere/FileSet/nvx021f54g
- "/opt/heracles/deploy/scholarsphere/releases/20170217172726/tmp/uploads/nvx021f54g_version1_RFP
  for Montgomery Burns Library System.rtf"
locale: en
Exception
URI::InvalidURIError
Error
bad URI(is not URI?): /tmp/nvx021f54g_version1_RFP for Montgomery Burns Library System.pdf
```

carolyncole:
This also seems to fix #661

carolyncole:
I have rerun several of the failing jobs after the code was deployed to QA.  None of those jobs continued to fail.   Need to run a full import to completely verify.

PR #688, opened 2017-03-03
Removing special characters from temporary file name

carolyncole:
refs #508 

This changes the file name from "abc123_version1_my label.pdf" to "abc123_version1_my_label.pdf" which should make the command line derivative tools happier.

### DIFF
--- a/lib/scholarsphere/import/version_builder.rb
+++ b/lib/scholarsphere/import/version_builder.rb
@@ -36,7 +36,7 @@ module Import
     private
 
       def create(file_set, version)
-        filename_on_disk = File.join Rails.root, "tmp/uploads", "#{file_set.id}_#{version[:label]}_#{file_set.label}"
+        filename_on_disk = temp_file_name(file_set, version)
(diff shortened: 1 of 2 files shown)


lib/scholarsphere/import/version_builder.rb
```diff
@@ -81,5 +81,10 @@ def characterize(file_set, filename_on_disk)
       def f3_to_f4_migration_date?(date)
         date.starts_with?("2015-04-11")
       end
+
+      def temp_file_name(file_set, version)
+        label = file_set.label.gsub(/[^0-9A-Za-z.\-]/, '_')
+        File.join Rails.root, "tmp/uploads", "#{file_set.id}_#{version[:label]}_#{label}"
```
awead, line 87: `Rails.root, "tmp/uploads"` is the same as `CurationConcerns.config.working_path`. We could reuse that, unless we want to define our own path separate from that.

awead: @cam156 I'm going to approve this so we can get it merged while you're away, but am filing #697 to keep the discussion.

Issue #590, opened 2017-02-02
Rename buttons in create collection page


carolyncole:
When a user does not have files in ScholarSphere
And they want to create a collection with files on disk or in Box
Then when they create a collection there is not really a clear path on how to upload into a collection

carolyncole:
"Create Collection" button should become "Create Empty Collection"
Add "Create Collection and Upload Works" -> This would take you to batch upload screen with collection set as the default relationship.
Add "Create Collection and Add Existing Works" -> This would take you to the dashboard where you can select works from the My Works List
 
<img width="840" alt="screen shot 2017-02-02 at 1 16 00 pm" src="https://cloud.githubusercontent.com/assets/1599081/22562262/c9755668-e949-11e6-86f3-3e4c9b58b695.png">

PR #695, opened 2017-03-07
Rename buttons in create collection page, ref #590

awead:
When creating a new collection, three different buttons will appear:
* creating a new, empty collection
* creating a new collection, then redirecting to the batch create page with the newly created collection pre-selected so that you may add new works to the collection.
* creating a new collection, then redirecting to the dashboard works page so that you can select existing works to add to the new collection.

new

Additionally, when doing the reverse of option 3 above, and selecting works from the dashboard first, then creating a new collection, only one button is shown. Also, a listing of the works that will be added to the collection is displayed using the same UI components that are used when displaying works in existing collections.

existing

Note: also includes some refactoring of the collection feature tests, separating create and delete tests.


mtribone (commented):
We have 3 buttons with the same ID `create_submit`. I'm guessing that it is for the JS?

awead:
@mtribone I'm not really sure. From what I can tell, the id isn't really used for anything. I'm happy to use different ids. Does this impact accessibility?

mtribone:
@awead I don't think that it will affect accessibility if the ID isn't being used for anything. It's a best practice to have only one ID with the same name on a page, but I wasn't sure if there was any JS using that ID when the button is clicked. For example, if `create_submit` kicks off any other processes. Not a blocker.

awead:
@mtribone the id isn't being used for anything, but I agree, probably better to use different ids. Let me update it.

Issue #640, opened 2017-02-10
Requirements icons do not read in screen reader need alt text


mtribone:
The check and x icons should have alt text to indicate to screen readers that they still have stuff to do.

carolyncole:
We need to change the java script to show hide actual icons and not after items in css.  The icons need alt text or real text so they are read.

mtribone:
@awead My thought was to hook onto the checks here: https://github.com/projecthydra/sufia/blob/7.2-migration/app/assets/javascripts/sufia/save_work/save_work_control.es6#L107

and inserting some text like:
https://github.com/psu-stewardship/scholarsphere/blob/8b603196665da8e7a0db769b8a1169d80b06f54a/app/assets/javascripts/scholarsphere/viz_widget.js

PR #698, opened 2017-03-09
Enable changing text for requirements, ref #640

awead:
Changes the text of the requirements links so that at screen reader informs the user as to whether the requirements are met.

Additionally, the links themselves need some Javascript reworking so they would active the tabs when clicked.

### DIFF
--- a/app/assets/javascripts/sufia/save_work/checklist_item.es6
+++ b/app/assets/javascripts/sufia/save_work/checklist_item.es6
@@ -1,0 +1,19 @@
+// Overrides Sufia checklist_item.es6 to add text for screen readers when metadata requirements are complete.
+
+export class ChecklistItem {
+  constructor(element) {
+    this.element = element
+  }
+
+  check() {
+    this.element.removeClass('incomplete')
+    this.element.addClass('complete')
+    this.element.children('a').text(this.element.data("complete"))
+  }
+
+  uncheck() {
+    this.element.removeClass('complete')
+    this.element.addClass('incomplete')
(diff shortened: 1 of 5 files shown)


awead:
@mtribone here's what I'm thinking. This changes the text of the link, however, the screen reader doesn't seem to pick it up, even with using aria. Maybe JAWS would?

Before:
before

After:
after

awead:
@mtribone  BTW, I initially tried a span, but it didn't seem to pick that up either.

mtribone:
@awead It's so simple it is wonderful. Tried to use the big JS hammer..._puts head into hands_

app/assets/javascripts/sufia/save_work/checklist_item.es6
```diff
@@ -0,0 +1,19 @@
...
+
+export class ChecklistItem {
+  constructor(element) {
+    this.element = element
+  }
+
+  check() {
+    this.element.removeClass('incomplete')
+    this.element.addClass('complete')
+    this.element.children('a').text(this.element.data("complete"))
```
mtribone, line 11: I tried to put the `<span>` here and it just repeated the content 5 times as I filled out the form.

awead:
@mtribone don't be hard on yourself 😉

awead:
@mtribone ok, this has changed a bit under the hood, but the concept is the same: change the links. I'm no longer concerned that the user hear the changes immediately, instead, the reader will announce the text when the user gets to it in the course of moving through the form.

Also, the links were updated with some JS to make them active the tabs.

PR #699, opened 2017-03-09
Make the Batch Upload Link More Apparent

mtribone:
Uses some JS to override and insert styles without overriding a template for a simple thing. Not super design sexy, but it works.

screen shot 2017-03-09 at 1 16 56 pm

### DIFF
--- a/app/assets/stylesheets/scholarsphere/form.scss
+++ b/app/assets/stylesheets/scholarsphere/form.scss
@@ -6,3 +6,7 @@
 html > body .form-progress .radio input[type="radio"] {
   margin-left: 0;
 }
+
+.switch-upload-type {
+  padding: 0.5em 0;
+}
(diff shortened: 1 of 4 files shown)


awead:
@mtribone interesting. I would have just overridden the view. Too much?

mtribone:
@awead If I do, I'll need to make note of it. Overriding for a class name has come up in the past. Go for it?

awead:
@mtribone meh. I'm good with this if you are. Could you put a comment in the layout.js to explain why it's there?

mtribone:
@awead I'm adding the overrides. :-)

mtribone:
screen shot 2017-03-09 at 1 58 22 pm

mtribone:
Overrides both screens to go back and forth between batch and single work.

Issue #674, opened 2017-02-24
Required metdata for collection


carolyncole:
Description and keyword should be required for collections

PR #701, opened 2017-03-10
Description and keyword are required for collections, ref #674

awead:


spec/forms/collection_form_spec.rb
```diff
@@ -18,4 +18,19 @@
...
   end
+
+  describe "#primary_terms" do
+    subject { form.primary_terms }
+    it { is_expected.to contain_exactly(:title, :description, :keyword) }
+  end
+
+  describe "#secondary_terms" do
+    subject { form.secondary_terms }
+    it { is_expected.not_to include(:title, :description, :keyword) }
```
carolyncole: @awead It would be better to check for exactly what we are expecting.  This test will allow any other terms including adding or removing existing terms.  We will want to know if that changes down the road.

spec/features/collection/create_spec.rb
```diff
@@ -29,6 +31,8 @@
     context "without any files" do
       it "creates an empty collection" do
         fill_in "Title", with: title
+        fill_in "Description", with: "description"
```
carolyncole: I wonder if these lines could be added into the before since they are set up and repeated for each test.

carolyncole (changes requested):

spec/forms/collection_form_spec.rb
awead: @cam156 these fields were under secondary terms prior to this change, hence the reason to ensure they are not there now, but I can check for the remaining terms as well.

carolyncole (approved):

Issue #643, opened 2017-02-10
Show Status or Feedback after saving a Work


mtribone:
A user creates a new work by adding the required metadata, visibility, and uploads files to be processed. Then checks the deposit agreement and hits save. 

System starts to process files, the save button is greyed out, but there is no indication that something is happening. Both sighted and users of screen readers asked for more feedback during this process, i.e., a status bar or percent complete.

carolyncole:
This should be read and visually shown.  "We are you saving your work this may take a few moments"  I think to get the screen reader to read this content on the fly the div will need to be labeled with aria-active.

PR #704, opened 2017-03-13
Use inline styles for progress bar, ref #643

awead:
@mtribone For whatever reason, Bootstrap needs inline styes here in order for the progress bar to show.

progress


mtribone:
Turds. Silly Bootstrap. I'll look away and hit the green button.

PR #710, opened 2017-03-17
Rename Labels on Work Tabs and File Set Edit

mtribone:
Adds content for tabs to sufia.en.yml and updates text. Changes views to use translation. Remove icons from tabs. Update tests to reflect name changes.

Changes based on user testing and accessibility testing with Christian and Michele.

### DIFF
--- a/app/views/curation_concerns/base/_guts4form.html.erb
+++ b/app/views/curation_concerns/base/_guts4form.html.erb
@@ -1,0 +1,47 @@
+<% # we will yield to content_for for each tab, e.g. :files_tab %>
+<% tabs ||= %w[metadata files relationships share] # default tab order %>
+<div class="row">
+  <div class="col-xs-12 col-sm-8" role="main">
+
+    <!-- Nav tabs -->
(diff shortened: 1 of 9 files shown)


mtribone:
screen shot 2017-03-17 at 3 01 00 pm

mtribone:
screen shot 2017-03-17 at 3 01 19 pm

mtribone:
screen shot 2017-03-17 at 3 01 55 pm

app/views/curation_concerns/base/_guts4form.html.erb
```diff
@@ -0,0 +1,47 @@
+<% # we will yield to content_for for each tab, e.g. :files_tab %>
```
carolyncole, line 1: @mtribone Did we need this entire override?  Was this to just change the tab name?

mtribone: It was to remove the icons. I may be able to use some CSS trickery to hide them, but I was able to remove them everyplace else except for the main form.

PR #711, opened 2017-03-20
Add Optional Steps to Save Viz Widget

mtribone:
Adds optional metadata links to save widget, adds tab names to JS file, and adds css for better spacing.

screen shot 2017-03-20 at 2 51 27 pm

### DIFF
--- a/spec/features/generic_work/upload_and_delete_spec.rb
+++ b/spec/features/generic_work/upload_and_delete_spec.rb
@@ -58,9 +58,13 @@ describe 'Generic File uploading and deletion:', type: :feature do
         click_link("Additional fields")
         expect(page).to have_no_css("#generic_work_contributor")
 
+        # Check for optional metadata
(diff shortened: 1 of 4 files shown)


carolyncole (changes requested):
@mtribone it would be nice to add a test (or add to a test to click on your new links).  That way if they get removed the tests would fail.  Otherwise looks good.

mtribone:
@cam156 👍

spec/features/generic_work/upload_and_delete_spec.rb
```diff
@@ -58,9 +58,13 @@
         click_link("Additional fields")
         expect(page).to have_no_css("#generic_work_contributor")
 
+        # Check for optional metadata
```
mtribone, line 61: @cam156 I've basically just added a click, but did not add any tests for content on each tab. Yes?

carolyncole (approved):

PR #715, opened 2017-03-23
Adding resolrize job that runs through all ids once and then send the…

carolyncole:
… models back again to make sure relationships are solrized

One thing to note we now need all descendents to capture the permissions.  Before with sufia 6 we just needed the top level descendents.

### DIFF
--- a/app/jobs/resolrize_job.rb
+++ b/app/jobs/resolrize_job.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+class ResolrizeJob < ActiveJob::Base
+  queue_as :resolrize
+
+  def perform
+    resource = Ldp::Resource::RdfSource.new(ActiveFedora.fedora.connection, ActiveFedora.fedora.host + ActiveFedora.fedora.base_path)
+    # GET could be slow if it's a big resource, we're using HEAD to avoid this problem,
+    # but this causes more requests to Fedora.
+    return [] unless resource.head.rdf_source?
+
+    descendant_uris = ActiveFedora::Base.descendant_uris(ActiveFedora.fedora.base_uri)
+
+    # map uri to id to make filtering types easier
+    descendant_ids = descendant_uris.map { |uri| ActiveFedora::Base.uri_to_id(uri) }.compact
+
+    # models have the 10 digit ids fron AF Noid so remove the longer ones
+    model_ids = descendant_ids.reject { |id| id.length > model_id_length }
+
+    # permissions have the longer ids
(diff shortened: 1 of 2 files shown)


app/jobs/resolrize_job.rb
```diff
@@ -0,0 +1,49 @@
...
+    resource = Ldp::Resource::RdfSource.new(ActiveFedora.fedora.connection, ActiveFedora.fedora.host + ActiveFedora.fedora.base_path)
+    # GET could be slow if it's a big resource, we're using HEAD to avoid this problem,
+    # but this causes more requests to Fedora.
+    return [] unless resource.head.rdf_source?
+
+    descendant_uris = ActiveFedora::Base.descendant_uris(ActiveFedora.fedora.base_uri)
+
+    # map uri to id to make filtering types easier
+    descendant_ids = descendant_uris.map { |uri| ActiveFedora::Base.uri_to_id(uri) }
+    descendant_ids = descendant_ids.uniq - [nil]
```
awead: adding `descendant_uris.map { |uri| ActiveFedora::Base.uri_to_id(uri) }.compact` should remove any nil items.

app/jobs/resolrize_job.rb
```diff
@@ -0,0 +1,49 @@
...
+    update_group(model_ids)
+  end
+
+  private
+
+    def model_id_length
+      10
+    end
+
+    def update_group(ids)
```
awead, line 38: Another option would be to create a second job, UpdateIndexJob, that does the actual reindexing. If there were any errors, they're get written to the Resque error queue. The advantage would be that multiple workers could run the updates, accomplishing things faster. Since you'd be pushing them on the queue in order, permissions first, then groups, we should be okay. A downside might be that the Resque error queue would get bigger and it might be harder to tell the indexing errors apart from the other errors.

spec/jobs/resolrize_job_spec.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+require 'rails_helper'
+require 'byebug'
```
awead: shouldn't need this.

awead (changes requested):

app/jobs/resolrize_job.rb
carolyncole: @awead we might end up with a timing issue depending on how the workers worked the jobs.  Aka jobs in different queues could be completed faster or slower.  It seems fairly safe, but at this point changing the update to be asynchronous is more that I would like to do.

awead (approved):

Issue #675, opened 2017-02-24
Making sure the index show only the fields we want


carolyncole:
What fields are those?

carolyncole:
remove Contributor, Date Published, and Date modified from index fields in the catalog controller, Only show the fields that are actual facet or sort values.

PR #719, opened 2017-03-23
Only display index fields that are actual facet or sort values, ref #675

awead:


app/services/field_configurator.rb
```diff
@@ -9,29 +9,36 @@ def self.common_fields
...
+      .except(:has_model)
+      .merge(depositor:  FieldConfig.new("Depositor"),
+             contributor: FieldConfig.new("Contributor"),
+             date_modified: FieldConfig.new(label: "Date Modified",
+                                            index_solr_type: :stored_sortable,
+                                            index_type: :date),
+             date_created: FieldConfig.new("Date Created"),
+             rights: FieldConfig.new("Rights"),
+             identifier: FieldConfig.new("Identifier"),
+             description: FieldConfig.new("Description"))
```
carolyncole, line 37: @awead isn't this going to put the description at the bottom of the page?

awead: @cam156 no, in fact, AFAICT, this doesn't affect anything relating to the show view because that's all done with the _attribute_rows.html.erb partial, and description is rendered separately in the _work_description partial.

carolyncole (approved):

PR #721, opened 2017-03-28
Removes the delete work link from the drop down.

mtribone:
This change removes the ability to delete a work from the edit page of a collection.

screen shot 2017-03-28 at 9 46 47 am

### DIFF
--- a/app/views/collections/_work_action_menu.html.erb
+++ b/app/views/collections/_work_action_menu.html.erb
@@ -1,0 +1,35 @@
+<!-- Removes the delete work option from the dropdown -->
+<div class="btn-group">
+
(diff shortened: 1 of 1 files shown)


app/views/collections/_work_action_menu.html.erb
```diff
@@ -0,0 +1,34 @@
+<div class="btn-group">
```
carolyncole, line 2: @mtribone can you add a comment to the top to indicate that we are adding the partial to remove the delete work option?

carolyncole (changes requested):

carolyncole (approved):

Issue #607, opened 2017-02-03
Missing Template error/500


carolyncole:
When I create a work 
And that creation errors
Then the error page should load instead of the server error page
![screen shot 2017-02-03 at 8 13 23 am](https://cloud.githubusercontent.com/assets/1599081/22592628/b0e55a22-e9e8-11e6-8c95-705cf81414c5.png)

```
ActionView::MissingTemplate (Missing template errors/500 with {:locale=>[:en], :formats=>[:html], :variants=>[], :handlers=>[:erb, :builder, :raw, :ruby, :coffee, :jbuilder, :haml]}. Searched in:
  * "/opt/heracles/deploy/scholarsphere/releases/20170130164732/app/views"
  * "/opt/heracles/deploy/scholarsphere/shared/bundle/ruby/2.3.0/gems/devise-4.2.0/app/views"
  * "/opt/heracles/deploy/scholarsphere/shared/bundle/ruby/2.3.0/bundler/gems/flipflop-0eecddade989/app/views"
  * "/opt/heracles/deploy/scholarsphere/shared/bundle/ruby/2.3.0/bundler/gems/sufia-603e8fc8bd4f/app/views"
  * "/opt/heracles/deploy/scholarsphere/shared/bundle/ruby/2.3.0/gems/browse-everything-0.10.5/app/views"
  * "/opt/heracles/deploy/scholarsphere/shared/bundle/ruby/2.3.0/gems/hydra-batch-edit-2.1.0/app/views"
  * "/opt/heracles/deploy/scholarsphere/shared/bundle/ruby/2.3.0/bundler/gems/curation_concerns-b073550c40a1/app/views"
  * "/opt/heracles/deploy/scholarsphere/shared/bundle/ruby/2.3.0/gems/qa-0.10.2/app/views"
  * "/opt/heracles/deploy/scholarsphere/shared/bundle/ruby/2.3.0/gems/hydra-editor-3.1.2/app/views"
  * "/opt/heracles/deploy/scholarsphere/shared/bundle/ruby/2.3.0/gems/blacklight-gallery-0.6.1/app/views"
  * "/opt/heracles/deploy/scholarsphere/shared/bundle/ruby/2.3.0/gems/blacklight_advanced_search-6.2.1/app/views"
  * "/opt/heracles/deploy/scholarsphere/shared/bundle/ruby/2.3.0/gems/blacklight-6.7.2/app/views"
[...]

PR #725, opened 2017-04-05
ErrorPresenter for customized error messages, ref #607

awead:
Returns a customized error page based on the class of exception raised.

We'll need an additional update of the i18n config file with the specific message desired for each exception.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -14,25 +14,26 @@ class ApplicationController < ActionController::Base
 
   before_action :clear_session_user, :filter_notify
 
-  rescue_from ActiveFedora::ObjectNotFoundError, with: :render_404 unless Rails.env.development?
+  rescue_from ActiveFedora::ObjectNotFoundError,
+              AbstractController::ActionNotFound,
+              ActionController::RoutingError,
+              ActionDispatch::Cookies::CookieOverflow,
+              ActionView::Template::Error,
+              ActiveRecord::RecordNotFound,
+              ActiveRecord::StatementInvalid,
+              Blacklight::Exceptions::ECONNREFUSED,
+              Blacklight::Exceptions::InvalidSolrID,
+              Errno::ECONNREFUSED,
+              NameError,
+              Net::LDAP::LdapError,
+              Redis::CannotConnectError,
+              RSolr::Error::Http,
+              Ldp::BadRequest,
+              StandardError,
+              RuntimeError, with: :render_error_page
 
-  unless Rails.env.development? || Rails.env.test?
-    rescue_from AbstractController::ActionNotFound, with: :render_404
-    rescue_from ActionController::RoutingError, with: :render_404
-    rescue_from ActionDispatch::Cookies::CookieOverflow, with: :render_500
-    rescue_from ActionView::Template::Error, with: :render_500
-    rescue_from ActiveRecord::RecordNotFound, with: :render_404
-    rescue_from ActiveRecord::StatementInvalid, with: :render_500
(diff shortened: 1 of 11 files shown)


app/controllers/application_controller.rb
```diff
@@ -14,25 +14,25 @@ class ApplicationController < ActionController::Base
...
+              ActiveRecord::StatementInvalid,
+              Blacklight::Exceptions::ECONNREFUSED,
+              Blacklight::Exceptions::InvalidSolrID,
+              Errno::ECONNREFUSED,
+              NameError,
+              Net::LDAP::LdapError,
+              Redis::CannotConnectError,
+              RSolr::Error::Http,
+              Ldp::BadRequest,
+              RuntimeError, with: :render_error_page
```
awead: I think we could also add `StandardError` to this so it would rescue anything else and always display the same error message page no matter what.

carolyncole: @awead That seems like a reasonable addition

app/controllers/application_controller.rb
```diff
@@ -47,15 +47,17 @@ def clear_session_user
...
   end
 
-  # Overrides CurationConcerns::ApplicationControllerBehavior with custom error method
+  # Overrides CurationConcerns::ApplicationControllerBehavior to use our error presenter
   def render_404(exception = nil)
-    log_exception(exception, "404") if exception.present?
-    render template: '/errors/404', layout: "error", formats: [:html], status: 404
+    @presenter = ErrorPresenter.new(exception)
+    @presenter.log_exception
+    render template: '/error', layout: "error", formats: [:html], status: "404"
```
carolyncole: @awead is there a reason you used `@presernter.status` below and `"404"` here?

config/locales/en.yml
```diff
@@ -56,3 +56,28 @@ en:
       subject: "ScholarSphere - Statistic Report"
       csv:
         file_name: "scholarsphere-stats_%{date_str}.csv"
+  errors:
```
carolyncole: @awead it would be good to add a comment here about how the keys are created since it seems non trivial.  It looks to me like the user is expected to take the exception class name and then convert it from Camel case and the substitute _ for the ::

carolyncole (approved):
If you want to make the change to add StandardError you could, but that would not necissarily block this change

app/controllers/application_controller.rb
awead: @cam156 cool. I'll add it in. I'm still not 100% convinced that any error we run into will get reported out this way, but meh, we'll see!

app/controllers/application_controller.rb
awead: @cam156  Because the render_404 method can have a nil exception, the presenter defaults to 500, but we need 404 here. I'm not a fan of a nil exception. I'd rather create a custom exception message, i.e. ScholarsphereError, which could have its own status (404), and message page, as well as a line in the log stating that, hey, we rendered 404 using this render_404 method, but we weren't given an actual exception so we're using this one to let you know we really don't know what happened. 😄

awead:
@cam156 Ok, I've spruced this up a bit. I've provided defaults for any exceptions that haven't been explicitly defined in the locales file, as well as written a log message that identifies missing keys from file should you want to add them.

If no exception is present, we use our default `Scholarsphere::Error` with its own custom message. And lastly, we're now rescuing from `StandardError` which catches *everything*, even `CanCan::AccessDenied` which has special, albeit strange, behaviors. So, I'm preserving the current behavior which is to redirect to login. We may want to revisit that in the future and maybe change it, but for now, everything should still behave the same.

PR #744, opened 2017-04-10
separating the config directories for qa and qa-new

carolyncole:
The old way  config_qa_new was a symbolic link to config_qa and all of the config files that changed between new qa and old qa were named with an "-new"

The new way the directories are separate:  config_qa and config_qa_new  and the file names remain the same.

This makes it easier to change the files in the config directory, but I just realized this does not document which files have been changed on qa-new server. @jrpatterson @informaticianme based on the lack of documentation I am not sure if this is a good or a bad change...

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -47,25 +47,25 @@ set :format_options, command_output: false
 # Example link: ln -s /opt/heracles/deploy/scholarsphere/shared/config/redis.yml /opt/heracles/deploy/scholarsphere/current/config/redis.yml
 set :linked_files, fetch(:linked_files, []).push(
   'config/analytics.yml',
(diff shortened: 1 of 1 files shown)


jrpatterson:
@cam156  Are you concerned you will not know which files to then update for other environments?  Since the current code explicitly defines what files were updated?

If so when the new config directories are created, it's best to copy with "cp -p" to preserve the timestamps then the files updated pretty much this year should be what is needing altered for the other environments.

carolyncole:
@jrpatterson I recreated the config_qa_new directory preserving the dates.  I also touched the *-new.yml files before I renamed them so everything that changed has been updated since April 10th.  DOes that mean we are good to go with this change?

PR #752, opened 2017-04-11
Move the directory linking to before the check of linked directories …

carolyncole:
…since it will fail if the directory does not exist

Realized doing it before the sym linking does it after the check that those directories and files exists.  Doing it on a new machine or a machine where the sym link does not exist in the first place causes the check to fail unless symlink_shared_directories is prior to deploy:check:linked_dirs


carolyncole:
@jrpatterson @informaticianme

Issue #730, opened 2017-04-05
Keywords are not keeping the input version AND not sorted by Alpha


carolyncole:
I added a regular key word and then all the letters as separate keywords in order.  When I got the record back out the keywords are scrambled
![screen shot 2017-04-05 at 10 37 26 am](https://cloud.githubusercontent.com/assets/1599081/24710678/e1d165e4-19eb-11e7-927f-262aa15735da.png)

mtribone:
Is it because keywords are a hash and not an array? I believe that hashes aren't concerned about order?

awead:
I don't believe hash keys or array elements are stored with an order. The keyword property is an array and doesn't have its order persisted. Maybe we should display them in alpha order always? It would be a small updated to the form and presenter or attribute rendered.

mtribone:
Hit it with a `.sort` method for alpha order?

mtribone:
I also recall us discussing this with @olendorf and he wasn't so concerned about the keywords being out of order and was more concerned with creator and contributor.

carolyncole:
This is also an issue with creator.

carolyncole:
We need to find out where the order is being lost.  Is it lost between the UI and fedora, or after it gets put into fedora.

Form and server look at parameters being posted.

carolyncole:
Here is the log from the form parameters going to the back end.  They are in order there: "creator"=>["Cole, Carolyn Ann", "a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l", "m", "n", "o", "p", "q", "r", "s", "t", "u", "v", "w", "x", "y", "z"],

carolyncole:
Here is the analogous issue occurring in  the import and how I fixed it https://github.com/psu-stewardship/scholarsphere/commit/9902f818079303db3e3ef8c468fff63ec289ffbc.  It seems very counter intuitive but if you assign the creator after the object has been saved, then when you save again the order stays.  If you do not reassign the order goes crazy...
https://gist.github.com/cam156/fb5acafa3833c07ba8c7b935f45ee013

PR #755, opened 2017-04-12
assigning the creator a second time to preserve order

carolyncole:
refs #730


carolyncole:
@awead @mtribone updated based on standup suggestions.   Also updated the test to save and reload the object to make sure we are looking at the item in fedora.

Issue #728, opened 2017-04-05
Demo videos in about page link should be removed


carolyncole:
This is related to #609.  The about page needs to link properly into the help page.

![screen shot 2017-04-05 at 10 09 34 am](https://cloud.githubusercontent.com/assets/1599081/24709466/0112cc80-19e8-11e7-9725-721c54209424.png)

PR #757, opened 2017-04-12
Removing unused static pages; refactor existing tests

awead:
This addresses #728, but as far as I can tell, the videos aren't there anymore. However, there are some other things that need cleaning up. There are some static pages that aren't used anymore: error_pages and subject_libraries. I can find no links to these pages and they aren't tested anywhere.

I've also edited the help page to remove the internal anchor links. The page is too short for them to be effective. I've also removed all of the commented-out html code.

The feature tests have been updated to reflect the existing pages.


app/views/static/help.html.erb
```diff
@@ -1,21 +1,9 @@
 <% provide :page_title, "Help | ScholarSphere" %>
 <h1 class="sr-only">Help</h1>
 
-<h2> Table of Contents</h2>
```
carolyncole: @awead This was added for accessibility.  We still have other sections so removing this make the page less accessible.

app/views/static/subject_libraries.html.erb
```diff
@@ -1,35 +0,0 @@
-<h1> Subject Libraries </h1>
```
carolyncole, line 1: @awead Have you looked at google scholar to see if this page is accessed?

carolyncole (changes requested):
Looks like you are removing an accessibility item.

app/views/static/subject_libraries.html.erb
awead: No, how do I do that?

app/views/static/help.html.erb
awead: Could we hide it then? It doesn't do anything for sighted users, other than reload the page.

awead: @cam156 I thought the headings themselves outlined the structure, which is what we're doing with the other pages, ensuring that headings move from H1, H2, etc. With the screen reader on, I can get an overview of the page just from the headings:
help

app/views/static/subject_libraries.html.erb
awead: @cam156 looks like it was removed at some point in QA: https://scholarsphere-qa.dlt.psu.edu/subject_libraries
It's still working in prod: https://scholarsphere.psu.edu/subject_libraries

awead:
@cam156 We agreed that subject_libraries could be removed. I put the TOC back into the help page.  Thanks!

carolyncole (approved):

Issue #729, opened 2017-04-05
Default Visibility is currently Private


carolyncole:
We used to default to public.  Guessing we still should

carolyncole:
We need to modify the default in work, batch and collection

carolyncole:
Make collections default to public

PR #760, opened 2017-04-12
Default visibility to public, ref #729

awead:
Defaults visibility to public at the form level.

Also includes refactoring feature tests to accommodate the change.


spec/features/generic_work/upload_and_delete_spec.rb
```diff
@@ -49,14 +49,11 @@
         within("#metadata")      { expect(page).to have_link("Licenses") }
         within("div#savewidget") { expect(page).to have_link("Required metadata complete") }
 
-        # Check for additional fields
-        expect(page).to have_no_css("#generic_work_contributor")
-        click_link("Additional fields")
```
carolyncole: @awead this used to check that the additional fields opened up when clicked.  Why would we want to remove that check?

spec/support/shared/forms.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
```
carolyncole, line 1: Cool!

carolyncole (changes requested):

spec/features/generic_work/upload_and_delete_spec.rb
awead: The animation wasn't working in the test so I took a different tack that we've using in other tests and don't test the animation. Instead, I'm testing that they're hidden.

PR #762, opened 2017-04-13
Update Solr to 6.4.1

awead:
This gets us to the current Solr version used in production and also gets all our solr configs in similar state.

### DIFF
--- a/config/travis/solr_wrapper_test.yml
+++ b/config/travis/solr_wrapper_test.yml
@@ -1,9 +1,9 @@
 #config/solr_wrapper_test.yml
-version: 5.5.2
+version: 6.4.1
 port: 8985
 instance_dir: solr-test
 download_dir: dep_cache
 collection:
-    persist: false
-    dir: solr/config
-    name: hydra-test
+  persist: false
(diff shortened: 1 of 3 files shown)


config/travis/solr_wrapper_test.yml
```diff
@@ -1,9 +1,9 @@
...
-version: 5.5.2
+version: 6.4.1
 port: 8985
 instance_dir: solr-test
 download_dir: dep_cache
 collection:
-    persist: false
-    dir: solr/config
-    name: hydra-test
+  persist: false
```
carolyncole, line 7: I may be wrong, but aren't these options for the collection?  So shouldn't thee spaces be there?

carolyncole (changes requested):
@awead have you been seeing any of the 503 errors locally based on this change?  I wonder if moving development to this version or changing the production version would be the best bet?

config/travis/solr_wrapper_test.yml
awead: @cam156 one or more spaces is sufficient. I was just using to for consistency.

Issue #738, opened 2017-04-06
Single Use Link does nothing


carolyncole:
When I am on a FileSet Page 
And I click the single use link button
Then something should happen

awead:
@cam156 are you not seeing the link created in the table? Or does there need to be more visual feedback?
![single](https://cloud.githubusercontent.com/assets/312085/24768692/8bf455dc-1ad1-11e7-85df-bfb91a8bbd07.png)

carolyncole:
@awead I only sometimes see the item in the table, 2 out of 4 times I hit the link an item got created
It also is below the fold on my screen so nothing seemed to have happened

carolyncole:
https://psu.box.com/s/1r4sv4gz188w3uucm98tvj2i61quz31e

awead:
I think the interaction ought to change so that the once the button is pressed, it's inactive until the row is added to the table, and then have a more noticeable UI change when it is added. Presently, it just seems there's a bit of a lag between the click and row insertion, and perhaps clicking again before the row has been loaded isn't registering.

carolyncole:
Disabling the button and possibly setting the table to active or moving the button to be in the same area as the table might be good so that a stupid user (aka me) notices the table and connects the two

carolyncole:
Need to figure out the accessibility of this code.  If it is not accessible this should be fixed within the ticket

PR #765, opened 2017-04-14
Making the single-use links more accessible, ref #738

awead:
Clicking the button to create a new single-use link wasn't making an obvious change.

The button was moved down under the table to which the links were being added, enabling the user to see the result more easily.
Additionally, aria labels and more informative text was added so that visually impaired links could more readily understand
what was occurring and what the links and buttons do.

The actions still aren't ideal. A better solution will be to not use tables and update the Javascript to include responses with
better accessibility functions.


carolyncole:
@awead can you include a screen capture?

awead:
@cam156 
sul

PR #767, opened 2017-04-18
Provide separate step for uploading all files, ref #453

awead:
Note: I'm not completely enamored with this solution because it basically side-steps the problem with a hack. However, I'm not really sure how else to fix the underlying problem because I think it lies with the jbuilder templating system and not with jQuery fileupload plugin. It also has a few hidden benefits, maybe? Uploads are actually occurring faster now, and there's a little more flexibility where the user can add files in multiple batches and only hit "Upload all" once.

This solution injects a new UI behavior where auto-uploading is disabled and a new button is provided to being uploading all the files after their records are individually added to the DOM. (see the circled button)

Although this does add an additional step to the workflow, it fixes the problem and also enables non-sequential, concurrent uploading for increased performance.
upload

ping @cam156 @mtribone @olendorf

### DIFF
--- a/app/assets/javascripts/sufia/uploader.js
+++ b/app/assets/javascripts/sufia/uploader.js
@@ -1,4 +1,9 @@
-// This is overriding Sufia's uploader.js so we can change the value of limitConcurrentUploads
+// Overrides Sufia's uploader.js
+// Injects new UI behavior where auto-uploading is disabled and a new button is provided to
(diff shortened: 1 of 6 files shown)


mtribone:
Well it may also give people the ability to cancel individual uploads in a "better" way, since I think before it was happening so fast on autoload that you had to be quick with the trigger.  So I'm guessing that a user can upload all OR individually upload items?

PR #774, opened 2017-04-20
Provide separate step for uploading all files, ref #453

awead:
Auto-uploading multiple small files was intermittently failing.

This solution injects a new UI behavior where auto-uploading is disabled and a new button is provided to being uploading all the files after their records are individually added to the DOM.

Additionally, the JS templates and other UI code are amended to provide better accessibility. This includes adding aria labels, appropriate table headers, and captions so screen readers can interpret the changes made to the table listing of files.

Although this does add an additional step to the workflow, it fixes the problem and also enables non-sequential, concurrent uploading for increased performance.

Watch: https://www.dropbox.com/s/sevs84srwzijrpk/uploading.m4v?dl=0

ping @cam156 @mtribone

### DIFF
--- a/app/assets/javascripts/sufia/uploader.js
+++ b/app/assets/javascripts/sufia/uploader.js
@@ -1,4 +1,9 @@
-// This is overriding Sufia's uploader.js so we can change the value of limitConcurrentUploads
+// Overrides Sufia's uploader.js
+// Injects new UI behavior where auto-uploading is disabled and a new button is provided to
+// being uploading all the files after their records are individually added to the DOM.
+// Although this does add on additional step to the workflow, it avoids the problems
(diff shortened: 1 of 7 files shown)


awead:
@cam156 @mtribone based on this morning's discussion, here's the updated result:
before:
before
after:
after
final:
final

I had to add an extra row with space to the right of the "all" button so we can show the progress of all the uploads. After it's done, the whole row and button go away.

awead:
@cam156 I'm not sure what to make of this. If I run the tests on the branch by itself, they pass:
https://travis-ci.org/psu-stewardship/scholarsphere/builds/225306200
The PR build, on the other hand, is failing consistently in the same place, but the place it's failing isn't any of the code that changed in the PR.

Issue #733, opened 2017-04-06
I can not save before all the files are done uploading


carolyncole:
When I choose a large number of files or ones that take a while to upload
Then the save button becomes active after the first file has uploaded
If I hit the save button before all the files have uploaded
Only the ones that have uploaded get added to the work

awead:
Fixed in Sufia https://github.com/projecthydra/sufia/commit/e61cb64322816485a191b69cdb7de2a07519379e

PR #784, opened 2017-04-25
Prevent saving a work when files are being uploaded, #733

awead:
The "Save" button is enabled immediately after the first file is uploaded when creating a work. However, if there is more than one file, the work could be save before the rest of the files have finished uploading.

We override Sufia's JS SaveWorkControl class so that the button is not made active until there no uploaded files in progress.


carolyncole (changes requested):
@awead Can you mark the location where you are changing the sufia base behavior?  It will make it easier to find it later.

awead:
@cam156 ha! I almost did that and thought, nah, that's too much.... keepin' me on my game, eh?

Issue #779, opened 2017-04-24
Remove the private option from the visibility


carolyncole:

awead:
I self-pointed this to 3

PR #786, opened 2017-04-26
Remove private visibility option, #779

awead:
Holding works in a private status will no longer be an option. Therefore, the option must be removed.

Existing private works will remain private and editing them will continue to allow them to be private even though the option is no longer explicitly present. New works can be embargoed, and collections will have no private option at all.

Note: I've tested this with existing works that are private. When editing them, there is no selected visibility, since it's private. However, if I save the work without selecting any visibility at all, it remains private. I would expect the form not to save at all since I hadn't selected a visibility. Maybe this is a bug that should be ticketed elsewhere.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ cache:
   bundler: true
   directories:
     - "dep_cache"
+    - "travis_phantomjs"
 bundler_args: --without development debug
 sudo: false
 rvm:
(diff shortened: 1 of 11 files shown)


.travis.yml
```diff
@@ -18,7 +19,17 @@ env:
     - TEST_SUITE=unit
     - TEST_SUITE=rubocop
 before_install:
-  - export PATH="$PATH:$(pwd)/fits"
```
carolyncole, line 21: Need this line?

PR #813, opened 2017-05-02
rescue any error that occurs deleting the file

carolyncole:
The error occurs whenever you restart a job and can mask other exceptions being thrown in other parts of the code.

### DIFF
--- a/app/jobs/import_version_job.rb
+++ b/app/jobs/import_version_job.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+
 class ImportVersionJob < ActiveJob::Base
   queue_as :files
   # @param [ActiveFedora::Base] the work class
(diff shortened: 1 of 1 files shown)


app/jobs/import_version_job.rb
```diff
@@ -7,6 +8,10 @@ def perform(file_set, filename_on_disk)
     # characterize the current version
     CharacterizeJob.perform_now(file_set, file_set.original_file.id, filename_on_disk)
   ensure
-    File.delete(filename_on_disk)
+    begin
+      File.delete(filename_on_disk)
+    rescue
```
awead: Should we rescue from StandardError and report the stack trace?

carolyncole:
@awead added the error message to the rescue logging, since we are rescuing one line I do not think we need a backtrace.

PR #831, opened 2017-05-04
Relabels Open Access to Public 

mtribone:
Overrides translation from Sufia and permission badge presenter from Curation Concerns. Rubocop fixes and updates spec test to replace open access with public. This change makes labeling more consistent in the interface and is more accurate with certain rights selections.

### DIFF
--- a/app/presenters/collection_presenter.rb
+++ b/app/presenters/collection_presenter.rb
@@ -4,4 +4,8 @@ class CollectionPresenter < Sufia::CollectionPresenter
   def self.terms
     super + [:date_modified, :date_uploaded]
   end
+
+  def permission_badge_class
+    PublicPermissionBadge
+  end
 end
--- a/app/presenters/file_set_presenter.rb
+++ b/app/presenters/file_set_presenter.rb
@@ -16,4 +16,8 @@ class FileSetPresenter < Sufia::FileSetPresenter
   def display_download_link?
     CurationConcerns.config.display_media_download_link
   end
+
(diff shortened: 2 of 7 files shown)


mtribone:
screen shot 2017-05-03 at 9 27 38 am

mtribone:
screen shot 2017-05-03 at 9 28 17 am

mtribone:
screen shot 2017-05-03 at 9 28 34 am

mtribone:
screen shot 2017-05-03 at 9 28 57 am

app/presenters/curation_concerns/permission_badge.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+module CurationConcerns
+  class PermissionBadge
```
awead: @mtribone can you create a new PermissionBadge class that inherits from CurationConcerns::PermissionBadge? That way you'll only have to override the link_title class. Then, you can override the `permission_badge_class` method in the file_set, work, and controller presenters, setting it to your new permission badge class. The badge class is assigned via CurationConcerns::PresentsAttributes which is a module that's included in different presenters. You end up having to touch more files, but I think it's a bit of cleaner override than having to change an entire file for just one line.

awead (changes requested):

mtribone:
Collection flavored Public

screen shot 2017-05-04 at 12 56 26 pm

Issue #783, opened 2017-04-25
Update the daily Audit to use the sufia audit


carolyncole:
https://github.com/psu-stewardship/scholarsphere/blob/develop/script/audit_repository

https://github.com/projecthydra/sufia/blob/7.2-migration/app/services/sufia/repository_audit_service.rb

awead:
@cam156 I am deeply suspicious of the Sufia:: RepositoryAuditService. If I forcibly corrupt a file in Fedora, the service doesn't report its failure.

awead:
@cam156 ok, it will recognize a failure. I forced it by removing all the records from ChecksumAuditLog. The failing file looks like:
```
#<ChecksumAuditLog id: 16, file_set_id: "83b5919121", file_id: "83b5919121/files/9eff14dd-9a61-4cdc-819a-006b1fa86...", version: "http://127.0.0.1:8986/rest/dev/83/b5/91/91/83b5919...", pass: 0, expected_result: nil, actual_result: nil, created_at: "2017-05-04 19:54:54", updated_at: "2017-05-04 19:54:54">
```
There's nothing for expected_result or actual_result, but at least it's catching the error.

PR #833, opened 2017-05-04
Update audit_repository script, ref #783

awead:
The previous script used a class that no longer existed.

We've updated the script to use the new Sufia service which reports errors in the same way as previous audits.

Failures are written to the ChecksumAuditLog files are only re-checked if the CurationConcerns.config.max_days_between_audits has passed since the last audit.


carolyncole:
@awead since there are no tests have you run this on the server or your local machine?

awead:
@cam156 local machine. You have to go into modeshape and corrupt a file

Issue #848, opened 2017-05-16
Use support email list for failed cron jobs


awead:
Cron job failures are currently being sent to UL-DLT-HYDRA@LISTS.PSU.EDU, which is defunct. We should use umg-up.its.scholarsphere-support@groups.ucs.psu.edu instead.

Issue #849, opened 2017-05-16
Refactor cronjobs


awead:
Cronjobs need some TLC. They seem to be shell scripts that call rake tasks, and sometimes report errors via email. Maybe we should have a consistent "interface" or process through which all jobs are run and make sure this is done in the same way for all the existing jobs.

PR #851, opened 2017-05-17
Update cronjobs with current emails, ref #848

awead:
Updates the emails used in cronjobs to send failure information. Also adds the daily stats jobs to schedule.rb and adjusts Capistrano's whenever settings so that crontabs are being cleared and regenerated correctly.

I'm not sure how send_daily_stats.bash was getting run unless it was managed somewhere else, so maybe that change needs to come out.

@cam156


awead:
There's more work to be done here regarding cronjobs, but that's targeted in #849

carolyncole (approved):

Issue #842, opened 2017-05-12
GA download stats will likely be broken


awead:
See https://github.com/projecthydra-labs/hyrax/issues/968

We don't appear to be using the `#file_download` selector so download click events will not get sent to Google Analytics.

Kudos to @hackmastera for finding this one!

awead:
I've confirmed on our QA system that download events are not being registered.

PR #852, opened 2017-05-17
Fix issue with capturing and rendering Google Analytics info, ref #842

awead:
GA was not registering any file downloads because the configured dom id wasn't there. Also, the rake tasks that were caching this information were not executing.

Since download links can be in multiple places, we're opting for a class that can be applied to any link and will register that event in Google. The rake task wasn't executing because existing loading mechanisms weren't including all the correct code from the Legato gem.

### DIFF
--- a/app/views/curation_concerns/file_sets/_actions.html.erb
+++ b/app/views/curation_concerns/file_sets/_actions.html.erb
@@ -1,0 +1,42 @@
+<%# Overrides Sufia to add class to download link to track statistics via GA %>
+<div class="btn-group">
+
+  <button class="btn btn-default dropdown-toggle" data-toggle="dropdown" type="button" id="dropdownMenu_<%= file_set.id %>" aria-haspopup="true">
+    <span class="sr-only">Press to </span>
+    Select an action
+    <span class="caret" aria-hidden="true"></span>
+  </button>
+
+  <ul role="menu" class="dropdown-menu dropdown-menu-right" aria-labelledby="dropdownMenu_<%= file_set.id %>">
+
+  <% if can?(:edit, file_set.id) %>
+    <li role="menuitem" tabindex="-1">
+      <%= link_to 'Edit', edit_polymorphic_path([main_app, file_set]),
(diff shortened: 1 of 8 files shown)


awead:
@cam156 This fixes the capturing of the stats and the display, but the data gathering has some problems.
stats
I've note one issue here #856 but I'm sure there are other issues that we should capture in different tickets.

app/views/curation_concerns/file_sets/_actions.html.erb
```diff
@@ -0,0 +1,42 @@
+<div class="btn-group">
```
carolyncole, line 2: @awead can you add a comment here indicating that you are overriding the file to allow for download tracking?

carolyncole (changes requested):

app/views/curation_concerns/file_sets/_actions.html.erb
awead: @cam156 it's on line 31, just above where the override is occurring.

carolyncole: @awead It is easier to find it when it is at the top of the file.  The other files had the change at the top.  Why in this one did you only add it on the line?

awead: @cam156 for the other files, it made more sense to put it near the `if` block, which coincided with the top of the file. In the case of this file, there are many `if can?` blocks and it's not immediately evident that we added an additional one. I wanted the comment to go closest the one that we were adding so that it was more evident what changes were being made.

carolyncole: @awead All I am asking for is an additional comment at the top of the file  to make it obvious the only reason this file was overridden was for download tracking.  Are you not willing to put that comment in?

awead: @cam156 I've moved the comment to the top of the file.

carolyncole (approved):

PR #858, opened 2017-05-18
Rake task to migrate Redis data to Sufia 7.2

awead:
I've tested this with production data copied locally, but it probably needs some more vetting.

### DIFF
--- a/lib/tasks/scholarsphere/redis.rake
+++ b/lib/tasks/scholarsphere/redis.rake
@@ -1,0 +1,28 @@
+namespace :scholarsphere do
+  namespace :redis do
+
+    desc "Migrates Redis keys to GenericWork types, removes SHARE events, and namespaces keys"
+    task migrate: :environment do
+      Redis.current.keys("*GenericFile*").each do |key|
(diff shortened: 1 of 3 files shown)


carolyncole (approved):

lib/tasks/scholarsphere/redis.rake
```diff
@@ -0,0 +1,20 @@
...
+
+      Redis.current.keys("*GenericFile*").each do |key|
+        length = Redis.current.llen(key)
+        Redis.current.lrange(key, 0, length).each do |id|
+          if Redis.current.hgetall("events:#{id}").fetch("action", "nothing").match(/SHARE/)
+            Redis.current.lrem(key, 0, id)
+            Redis.current.del("events:#{id}")
+          end
+        end
+        Redis.current.rename(key, key.gsub(/GenericFile/, "GenericWork"))
```
carolyncole: @awead does this need to be sufia:

awead: Ah, yes, I believe you're right.

Issue #853, opened 2017-05-17
Redis Events where do they go?


carolyncole:
Events in the new system are attached at the file level.  Events in the new system are attached at the GenericWork and the FileSet.   How do we want to translate the old system events into the new system.

Here is a screen shot of the events in the old system
![screen shot 2017-05-17 at 1 58 10 pm](https://cloud.githubusercontent.com/assets/1599081/26168690/022f7126-3b09-11e7-9674-5a1c941f2ad3.png)

awead:
FYI: there are just over 171K events in the dump from stage. 169K are for SHARE notify. 157K of them state "File could not be sent to SHARE Notify"

awead:
On the new system, there are events for creating and updating works which get attached to the Work object, as well as creating and updated files, which are attached to the FileSet object. Additionally, there are events for creating and restoring versions of files which are also attached to the FileSet.

However, _we are not displaying the Work events on its show page_, only FileSet events on its show page under User Activity. The only place that work-level events get shown is under User Activity in the dashboard, but there both work and file set events are interleaved together.

Here are some options:
1. Move all generic file events from the old system to FileSet events in the new system. This would mean no UI changes, but we'd have to create new entries for the file sets that are created during migration and move the old generic file events to the new file sets.

2. Move all generic file events from the old system to Work events in the new system. Here we add some UI elements to the work show page and move the old events right over. We don't have to create new entries because the GF ids map to Work ids, however, the events for new file sets created during the migration would show no activity.

3. Some combination of 1 and 2 where we "make up" some events for file sets that show when they were created or updated. I'm not sure how possible that is.

(Optional) prune all the SHARE Notify events. They're kind of noise right now, and not very usefull since we've disabled it and will need to re-send everything later.

I'm leaning towards option 2. We'll be showing the presently-hidden work events, and the migration is a bit easier. I think it's acceptable to have empty activities for migrated files, but that's just me!

olendorf:
I think option two is better as well, there isn't a perfect solution here though.

I think pruning the SHARE notify is almost a necessity. I think it is noise and not very useful as Adam points out. If we are presenting it to the user, it just makes them confused, or ignore it all together.

carolyncole:
Add the view to the work to display the activity view on the work.

PR #871, opened 2017-05-23
Display events for works, ref #853

awead:
Sufia captures events that occur on works, but they were not being displayed in the UI.

This displays the same partial that we use for file set events, only the events are attached to the work presenter.

### DIFF
--- a/spec/presenters/work_show_presenter_spec.rb
+++ b/spec/presenters/work_show_presenter_spec.rb
@@ -61,4 +61,23 @@ describe WorkShowPresenter do
     subject { presenter.facet_mapping(:creator) }
     it { is_expected.to eq("JOE SMITH" => "Joe Smith") }
   end
+
+  describe "#events" do
+    context "with no events" do
+      let(:work)   { build(:work) }
+      its(:events) { is_expected.to be_empty }
+    end
+
+    context "with events" do
+      let(:events) { double }
+      before do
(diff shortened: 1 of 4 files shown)


spec/presenters/work_show_presenter_spec.rb
```diff
@@ -61,4 +61,13 @@
     subject { presenter.facet_mapping(:creator) }
     it { is_expected.to eq("JOE SMITH" => "Joe Smith") }
   end
+
+  describe "#events" do
+    context "with no events" do
+      let(:work)   { build(:work) }
+      its(:events) { is_expected.to be_empty }
+    end
```
carolyncole, line 69: @awead can you write a test for a presenter with events?

carolyncole (changes requested):

spec/presenters/work_show_presenter_spec.rb
awead: @cam156 I could, but it would involve a lot of mocking of RedisEventStore. The events method there looks like `event_store.for(stream[:event]).fetch(size)`. The feature test is checking for the presence of events with actual data from Redis, so I felt we had it covered pretty well.

awead: @cam156 I pushed one up anyway... I'm not sure it's completely useful, but I guess is does check the key.

carolyncole (approved):

Issue #585, opened 2017-01-31
Create a read only mode for ScholarSphere 3.0


carolyncole:
As a ScholarSphere administrator I want to turn off all uploads and modifications so I can perform system maintenance.

When a configuration flag is set 
then has_access? shouls always display a readonly page 
Things that need to be done
* Create read_only flag in configuration
* create read_onlu maintenance page to be displayed
* check flag in has_access (https://github.com/psu-stewardship/scholarsphere/blob/master/app/controllers/application_controller.rb#L85)

carolyncole:
Grab this commit and update it for ScholarSphere 3.0 https://github.com/psu-stewardship/scholarsphere/commit/a586ae65325acc1340e8d6bd667a6e19bef5694e

Also need to update the migration documentation for how to set the 3.0 system to read only if the flag changes or the location of where it is getting set.

PR #879, opened 2017-05-25
Adding the ability to set the system to read only mode

carolyncole:
refs #585


config/application.rb
```diff
@@ -93,5 +93,8 @@ class Application < Rails::Application
 
     config.action_mailer.default_options = { from: "umg-up.its.sas.scholarsphere-email@groups.ucs.psu.edu" }
     config.action_mailer.default_url_options = { host: config.service_instance, protocol: 'https' }
+
+    # Set the  system to read only mode.  Does not allow new uploads, file edits, new collections, and collection edits
+    config.read_only = false
```
awead: @cam156 you could use our application.yml file and do `ENV.fetch("read_only", false)` that way we should be able to invoke read-only mode by just changing the yaml file on the server instead of the application code... but either way, this looks good to me.

carolyncole: @awead good idea I will change that.

awead:
@cam156 can you squash these? I was going to do it via GitHub, but I'm wary...

carolyncole:
@awead I did the conflict resolution in github just for grins and giggles.  I will go back and do it from the command line.

awead:
@cam156 huh. never done that before...

Issue #895, opened 2017-05-31
FiletSet depositor is incorrect when uploading on behalf of


awead:
When uploading a work on behalf of another user, the work's permissions have depositor as the "on behalf of" user, and the person uploading has edit access. However, the file set's permissions are reversed: the depositor is the person uploading while the "on behalf of user" has edit access.

awead:
There's a three-way race condition between `GenericWorkActor`, `AttachFilesToWorkJob` and `ContentDepositorChangeEventJob`. `AttachFilesToWorkJob` uses the work's depositor as the depositor for the file set. However, which depositor is set on the work depends on when `ContentDepositorChangeEventJob` has run.

The first time `apply_depositor_metadata` is called on the work is in `GenericWorkActor` via `CurationConcerns::Actors::BaseActor`, where it is set to the current user, *not* the `on_behalf_of` user.

At some point later, `ContentDepositorChangeEventJob` runs which changes the value of the depositor to the `on_behalf_of` user. But, note, that at this point there are no file sets attached so even though the job would also change the file set depositor in a transfer situation, is does not here.

When we get to `AttachFilesToWorkJob`, it's looking at the work's depositor.  If `ContentDepositorChangeEventJob` has run, it will be the correct `on_behalf_of` user. This is the case when running locally and all jobs are run inline, and it also seems to be the case when running with Resque locally because the actors apparently execute fast enough to invoke `ContentDepositorChangeEventJob` before `AttachFilesToWorkJob` gets called.

However, in production deployments to staging and QA, it appears that `AttachFilesToWorkJob` wins the race and attaches the file set to the work before `ContentDepositorChangeEventJob` has a chance to finish.

A possible solution is to override `apply_depositor_metadata` in our local `GenericWorkActor` to ensure the depositor will always be the right one regardless of what `ContentDepositorChangeEventJob` does.

carolyncole:
Still needs to be fixed based on research done: #978

PR #922, opened 2017-06-08
Adds the proxied user as the depositor in BaseActor, #895

awead:
Due to a race condition between GenericWorkActor, AttachFilesToWorkJob and ContentDepositorChangeEventJob, the file set's depositor could be set at the proxy user instead of the proxied user, that is, the user whom is being deposited on behalf of.

This addresses the problem by ensuring the work is set to the proxied user at the first opportunity. ContentDepositorChangeEventJob will still run, which also sets the depositor to the proxied user, but doing it in the actor first ensure that when AttachFilesToWorkJob runs, it will always have the correct depositor set.

Note: I'm not crazy about this solution. Particularly because it doesn't really address the underly problem. Also, the test isn't really covering it because you can't replicate the problem in test since all the jobs and actors are run inline. Still, it ensures that the code is at least executed.

### DIFF
--- a/app/actors/curation_concerns/actors/generic_work_actor.rb
+++ b/app/actors/curation_concerns/actors/generic_work_actor.rb
@@ -4,8 +4,13 @@
 module CurationConcerns
   module Actors
     class GenericWorkActor < CurationConcerns::Actors::BaseActor
+      # Override CurationConcerns to pass attributes to apply_creation_data_to_curation_concern
+      # and re-apply creator to ensure order is correct.
       def create(attributes)
-        stat = super(attributes)
+        @cloud_resources = attributes.delete(:cloud_resources.to_s)
+        apply_creation_data_to_curation_concern(attributes)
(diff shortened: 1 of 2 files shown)


carolyncole (commented):
Looks good.  I thought the issue was in the fileset depositor.  Does setting it here copy it down to the fileset?

awead:
@cam156 yes. It happens here:
https://github.com/psu-stewardship/scholarsphere/blob/develop/app/jobs/attach_files_to_work_job.rb#L11
where it gets the current depositor on the work, which is then passed to the new file set. Herein lies the race condition. If ContentDepositorChangeEventJob has already run, this depositor will be the proxied user. If it is hasn't it'll be the other (incorrect) user. So this is a kind of "head them off at the pass" solution. Ideally, (?) we could ensure ContentDepositorChangeEventJob always runs before this jobs does, but that's probably more complicated.

Issue #934, opened 2017-06-12
Analytics page is throwing an error on every record


carolyncole:
As described by Jeffrey Edmunds" <jhe2@psu.edu> the analytics page is just erroring

To reproduce this pick any work of files set on scholarsphere.psu.edu and click on the analytics button

carolyncole:
The page is not showing correct information, but at least it is not showing an error after I updated the configs for ticket #929

It looks like for migrated files the migration date is showing not the deposited date

mtribone:
I've also found that the @stats.created_date in the labeling at the top is having an issue with string to date.

https://github.com/samvera/sufia/blob/f189645d84bc2c796df444ed0b85d8ca84bc84c2/app/presenters/sufia/stats_usage_presenter.rb#L29

mtribone:
![Screen Shot 2017-06-14 at 8.39.26 AM.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5534fbcc6e7e601a00315e94/125516c66e82c728ace21e0d46db978826878de9bce1b057e40fa0c57e0b77256455f76b7191a370f111224f5d595edd3c4a4356bcbbd671f3e57862804e0ee984390975933943aabb821bb700b3208b4982b0bcdc4bd2f42f67b6d5db1063474f4dcbed5727a39b41c95cff4ce54951a4173561e6762431.png)

carolyncole:
https://gist.github.com/cam156/613af195d8a8a7b55d6559db84a229db

carolyncole:
ref #842

GenericFile View should translate into a work View 
GenericFile Download should translate into a FileSet Download.

The user should be able to see total views on a work, total views for a file set, and total downloads for a file set.

carolyncole:
The Data is in the wrong table and will not be displayed.

Any Items in the workview table prior to june 8th, 2017 are bogus and should be removed

We need to move all FileViews prior to June 8th, 2017 into the WorkView table transforming views column to work_views column, and the file_id column to  the work_id column.

carolyncole:
Saved the original tables
```
create table work_view_stats_sav as select * from work_view_stats;
create table file_view_stats_sav as select * from file_view_stats;
```
tested 1 file in the mysql console with these commands:
```
delete from work_view_stats where work_id = 'p8418n32c' and date < '2017-06-08';
insert into work_view_stats select * from file_view_stats where file_id = 'p8418n32c' and date < '2017-06-08';
[...]

PR #940, opened 2017-06-14
Convert dates to strings in Sufia::StatsUsagePresenter, ref #934

awead:
string_to_date was returning nil for parameters that were already DateTime objects because
it does not implicitly convert them to strings.

To fix, we explicitly call to_s to handle the conversion.

@cam156 this is a new technique I was reading about from @jcoyne.

### DIFF
--- a/app/prepend/prepended_stats_usage_behavior.rb
+++ b/app/prepend/prepended_stats_usage_behavior.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+# Using Rails' to_prepare hook in config/application.rb, we can inject this module as
+# Sufia::StatsUsagePresenter is loaded, and alter its existing behaviors without having to redefine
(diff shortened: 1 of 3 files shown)


carolyncole:
@awead it looks like there are too many commits on this PR.

carolyncole:
@awead interesting rails concept.  Are we following the rails pattern by putting the prepend in the app directory?  This is close to a monkey patch, so should it live in initializers?

awead:
@cam156 it's kind of like monkey patching, but it's something that's built into Rails. See point number 6 here: http://guides.rubyonrails.org/configuring.html 

Since it's a configuration option, it resides in application.rb. The code that's being prepended can be anywhere, I suppose, so long as it's loaded before the `to_prepend`, which should be just about anywhere.

Issue #949, opened 2017-06-16
Creator order should be preserved in metadata not just implied


carolyncole:
ref #369

We should have a way of preserving the order that creators were input into the system that does not rely on RDF 1.0 semi ordering.

This could be a secondary display field, or multiple creator fields by number or something else.

There are no changes to the user form, we are just making sure we capture the order in a more permanent way.

awead:
Here's how UVA did it: https://github.com/uvalib/libra-oc/commit/5d88527fcd68ef27a45aaf4f68bb23624d000d85

Later, they refactored it into a module https://github.com/uvalib/libra-oc/commit/0c9be010e344112eb8573ac7af1a475df66be3cf

mtribone:
Are we liking the way that UVA handled it?

awead:
One thing to note is that is looks like they are not using an additional predicate. They are serializing/deserializing the values using one predicate with a delimiter. One drawback to this is that it's not very linked-data-like. That's not really an issue since we aren't publishing any of this information; however, if we were to, we'd the to change this implementation so that the values are persisted using the right predicate. We'd want to use dc.creator for the real values, and have those values published, but then serialize/deserialize them as ordered elements in a different predicate, something like psu.orderedCreator, and definitely *not* publish that predicate.

mtribone:
So do we want to "future  proof" it and go with a different solution? Like the title and subtile solution and use different fields with different predicates?

awead:
I'm leaning towards the two-predicate solution. However, I think @olendorf and @kestlund should weigh-in on this.

olendorf:
@awead Good argument for two predicates, I say do it the right way.

mtribone:
So there is a predicate for `creator`, but is there a predicate for co-creator?
https://github.com/ruby-rdf/rdf-vocab/blob/develop/lib/rdf/vocab/ebucore.rb#L1523

mtribone:
Specialization or Ordinal Properties?

http://infolab.stanford.edu/~stefan/daml/order.html#fig:specialization
http://infolab.stanford.edu/~stefan/daml/order.html#fig:ordinalprop

awead:
I chatted @kestlund on Slack about this and her vote is _I would do all as dc.creator, then use a private ordering predicate or mechanism._

mtribone:
@awead What would be an example of a private ordering predicate or mechanism? Do the two that I referenced above apply? Just trying to understand a little more.

[...]

Issue #948, opened 2017-06-16
Title should be singular with subtitle


carolyncole:
refs #369

We need to change the work form and collection form to allow for only one title.

We need to add sigular subtitle to work and collection.
This ticket does not include doing this for file set.

carolyncole:
If there are two titles change the second title to subtitle field.  If there are more than two the last title get moved into the subtitle field separated by a semi colin

mtribone:
There is a solution in Hyrax that might be useful.

http://samvera.github.io/customize-metadata-other-customizations.html#making-a-default-property-single-value

PR #964, opened 2017-06-26
Proxied user is the depositor in BaseActor, #895

awead:
This is a reattempt of the solution that was first tried (unsuccessfully) in f5cadae53e887df9292afabe4ca950e87412ae6b.

It also refactors the actor code slightly to account for future changes that will need to occur when #948 and #949 are closed.


awead:
I pushed this up to QA and it looks good.

Issue #958, opened 2017-06-19
Uploading Multiple Files from Dropbox Gives Not Found Error


mtribone:
When I create a new work or edit an existing work and add multiple files from Dropbox, I get a "not found" error after i hit save.

![screen shot 2017-06-19 at 10 42 40 am](https://user-images.githubusercontent.com/4163828/27291007-f2798486-54dc-11e7-88a1-099a28bb9fc1.png)

carolyncole:
[...]

PR #965, opened 2017-06-27
Uploading in edit mode via BrowseEverything, #958

awead:
Adding new files to existing works via BrowseEverything was failing because the form was not using the correct target selector. As a result, BE would insert the wrong html into the form and files could not be uploaded.

This uses a new form method to determine the selector in each case, new and edit views for both single works and batches.


awead:
@mtribone can you review this? The key change is putting the logic for determining a DOM id in the form. I did this because we need to generate an id based on the work's own id, and the form know that. It also makes the view simpler.

PR #971, opened 2017-06-29
Multiple file sets with different permissions

awead:
This enables the display of works with different files sets of differing permissions, if the representative is of a lesser visibility. For example, a non-public representative for a public work. There's some test refactoring as well, and I've also resigned the edit form slightly with headings for each section: required metadata, media, and additional metadata. @mtribone should have a look to check accessibility issues. Here's what the new form looks like:
media


carolyncole:
@awead @mtribone was this reviewed for accessibility?

mtribone:
@cam156 There are accessibility issues with that form, according to the WAVE toolbar, that _aren't_ from @awead's code changes and updates for representative media and the thumbnail. I'd suggest other tickets for those issues.

Issue #886, opened 2017-05-26
Adding/Removing from others' collections


awead:
If you're an admin, you can add files to someone else's collection, but you can never remove them. Hotel California!

carolyncole:
This is really that the remove from collection button does not appear if there is only one item in the colection
![screen shot 2017-06-15 at 2 08 14 pm](https://user-images.githubusercontent.com/1599081/27195300-321e2300-51d4-11e7-94ed-3c92021bd7b4.png)

carolyncole:
![screen shot 2017-06-19 at 9 30 13 am](https://user-images.githubusercontent.com/1599081/27287514-f3994c6c-54d1-11e7-85bd-cc29181584c5.png)

carolyncole:
Put the button back when there is only one item.

carolyncole:
This should be ticketed in hyrax when this item is complete

PR #973, opened 2017-07-06
Render "Remove from Collection" button, ref #886

awead:
The button that removes an item from a collection was not rendered if the collection only had one item. Updating the view allows the button to be used no matter how many items are in the collection.

### DIFF
--- a/spec/views/collections/_sort_and_per_page.html.erb_spec.rb
+++ b/spec/views/collections/_sort_and_per_page.html.erb_spec.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+describe "collections/_sort_and_per_page.html.erb" do
+  let(:collection) { build(:collection, id: "coll1") }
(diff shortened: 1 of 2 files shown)


spec/views/collections/_sort_and_per_page.html.erb_spec.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+describe "collections/_sort_and_per_page.html.erb" do
+  let(:collection)    { build(:collection, id: "coll1") }
+  let(:solr_response) { Blacklight::Solr::Response.new({ response: { numFound: 1 } }, nil) }
```
carolyncole: @awead I do not think you need this let here.

carolyncole (changes requested):
@awead is this also showing the sort when there is only one item?  That may be a bit confusing.

awead:
@cam156 sort options are only shown when there's more than one work in the collection. The test has also been updated.

carolyncole (approved):

PR #976, opened 2017-07-07
Patch AttributeRenderer

awead:
Fixes two issues: 1) an update to I18n was breaking microdata, and 2) moves a previous method override to a prepend for better clarity.

Note: the second fix has its unit test covered in `translated_facet_attribute_rendere_spec`


app/prepends/prepended_renderers/configured_microdata.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+# Overrides CurationConcerns::Renderers::AttributeRenderer so we can correctly determines
+# if a key exists instead of attempting to return a false value because I18n no longer
+# interprets these as booleans.
+module PrependedRenderers
+  module ConfiguredMicrodata
+    def microdata?(field)
+      return false unless CurationConcerns.config.display_microdata
+      key = "curation_concerns.schema_org.#{field}.property"
+      t(key) if I18n.exists?(key)
```
carolyncole, line 10: @awead Should this return the boolean for exists?

carolyncole (changes requested):

app/prepends/prepended_renderers/configured_microdata.rb
awead: @cam156 no, it returns nil if the key doesn't exist.

carolyncole: @awead I would assume that a ? method would return a boolean value, either true or false, not the actual value or nil.  Is there a reason for this change?

awead: @cam156 it's in the original method: https://github.com/samvera/curation_concerns/blob/master/app/renderers/curation_concerns/renderers/configured_microdata.rb#L4-L8
In this case, `microdata?` returns the key, which is odd, but that that's the behavior, I guess. A different, but similar behavior exists in hyrax too: https://github.com/samvera/hyrax/blob/master/app/renderers/hyrax/renderers/configured_microdata.rb#L6

carolyncole: @awead I'm concerned we are propagating a bug.  From bgrep the only place this method is called is: https://github.com/samvera/hyrax/blob/master/app/renderers/hyrax/renderers/configured_microdata.rb#L31

Which darn well looks like it is expecting a boolean.

awead: @cam156 `AttributeRenderer` is making use of `ConfiguredMicrodata`'s `microdata_object_attributes` and `microdata_value_attributes`, which in turn both make use of `microdata?` and `microdata_object?`. These both return false or values, instead of just true/false. We could override the entire module and change these method names, but that seems like more extra work.

I'm less inclined to open up the whole module if I can make smaller change. The methods aren't exactly kosher, but they're not really egregious either.

carolyncole: @awead I am not suggesting changing the method name.  I am suggesting changing the code to return a boolean.  The name implies a boolean.  Where it is used implies it is expecting a boolean. All I was looking for was the code to change  to `I18n.exists?(key)` on the line.  It seems silly to continue to propagate what looks like a copy paste error.

PR #982, opened 2017-07-11
Removes File Format Metadata for Works from Search Index View 

mtribone:
Removes the file format field on the search view for works from the file configurator.

Removes file format from the common fields but adds it to the facet fields in the field configurator. Removes file format spec test from catalog controller, works controller, and field configurator because we are dealing with works not files.

### DIFF
--- a/spec/controllers/catalog_controller_spec.rb
+++ b/spec/controllers/catalog_controller_spec.rb
@@ -27,7 +27,6 @@ describe CatalogController, type: :controller do
       subject { described_class.blacklight_config.index_fields.keys }
       it { is_expected.to contain_exactly("based_near_tesim",
                                           "date_uploaded_dtsi",
-                                          "file_format_tesim",
                                           "keyword_tesim",
                                           "language_tesim",
                                           "publisher_tesim",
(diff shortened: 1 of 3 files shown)


mtribone:
screen shot 2017-07-11 at 9 26 01 am

spec/controllers/catalog_controller_spec.rb
```diff
@@ -134,13 +133,6 @@
         expect(assigns(:document_list).count).to be(1)
         expect(assigns(:document_list)[0].fetch(solr_field("resource_type"))[0]).to eql('resource_typeresource_type')
       end
-      it "finds a file by format_label" do
```
carolyncole: @mtribone does this test fail?  Since you are still including the facet I would expect this test to pass

carolyncole (changes requested):

spec/controllers/catalog_controller_spec.rb
mtribone: @cam156 Yes this test was failing until I removed it.

mtribone: https://travis-ci.org/psu-stewardship/scholarsphere/jobs/252117848#L2250

carolyncole: @mtribone since you kept the facet that test "Should" pass.  There may need to be a tweak to the test now that the facet is not displayed the results.

carolyncole (approved):

spec/controllers/my/works_controller_spec.rb
```diff
@@ -134,13 +134,6 @@
         expect(assigns(:document_list).count).to eql(1)
         expect(assigns(:document_list)[0].fetch(solr_field("resource_type"))[0]).to eql('resource_typeresource_type')
       end
-      it "finds a file by format_label" do
```
carolyncole: @mtribone I thought this was going to be back in.

carolyncole (changes requested):

spec/controllers/my/works_controller_spec.rb
mtribone: @cam156 Oops. forgot about the works controller

carolyncole (approved):

PR #988, opened 2017-07-13
Only show the representative media section when editing a work

mtribone:
This fixes the media section appearing on the new work page when the media hasn't been set. Only display the media section with values when we are editing a work.  Uses if action == edit and tests that the media section is not present on the new work form.


mtribone:
New work view

screen shot 2017-07-13 at 10 30 24 am

mtribone:
Edit work view

screen shot 2017-07-13 at 10 30 55 am

carolyncole (approved):

Issue #892, opened 2017-05-30
Multiple file sets and permissions


awead:
If a public work has multiple file sets, there will be an error when displaying it if the representative file set is not public as well.

To reproduce:
1. Create a public work with two file sets.
1. In another browser, where you are not logged in, view the public work
1. Edit the work's representative file set to have Penn State visibility
1. Reload the work in the other browser

carolyncole:
1) Let them choose any representative, but warn if the representative is PSU, embargoed, or private.

2) Show the default icon if the representative is not available (aka act as if there is no thumbnail).

carolyncole:
Seems to work on the work show, but the index still shows a broken link

![Screen Shot 2017-07-14 at 9.21.27 AM.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/54e5f67e04dc270b004037e9/125516c66e82c728ace21e0d46db978826878de9bce1b057e40fa0c57e0b77246455f76b7191a371f1102a4f5d585edd3c4a4356bcbbd671f3e57862804e0ee984390975933943aabb821bb700b3208b4982b0bcdc4bd2f42f67b6d5db1063474f4dcbed5727a29f41c95df64be14e5da1123f66e7752431.png)

awead:
@cam156 @olendorf this is opening up a can of worms now. The show view works because we're using presenters. In the search results, we use a thumbnail field in the solr response. In the above case, the image is broken because the user has specified a restricted image which which the public user cannot see. The application (correctly) returns a 401 and there's a broken image.

If we would like restricted thumbnails to show in the search results for authorized users, and show something else for unauthorized users, we have three options:
1. Query solr for each search result to determine if the user has read access to the representative file set
2. Capture the 401 response in the controller and return a default image for thumbnails
3. Disguise the broken image using some Javascript that just replaces the image with a default one if there's an error.

The first option isn't ideal because you're adding an additional solr query per search result, which is going to 10 for each page, and more if you show more results. The second option is okay, but is a bit invasive and requires the most code. That brings us to option three, which is the least invasive and requires the least code (about 4 lines).

I'm leaning towards option three at the moment; however, another idea is simply disallow users to specify thumbnails and representatives that don't have the same level of visibility. That would eliminate the problem altogether. The only additional step would be to catch changes to exiting representatives, and if their visibility becomes more restricted than the work, the work's thumbnail in Solr gets set to the default.

[...]

PR #1003, opened 2017-07-19
Replace broken thumbnail images with defaults, ref #892

awead:
This is the "option three" solution. See my notes to #892. It should only apply to images within the catalog display, which are search results. Any other image would appear with the standard "?" in the blue box or broken image default.


carolyncole (commented):
This seems a bit strange to me.  How did you handle the issue for Work Show and Edit?  Similar code?

awead:
@cam156 those use null presenters.

Issue #1001, opened 2017-07-17
Clean up depreciation warning when running rspec


carolyncole:
When we run rspec there are a number of deprecation warnings and a warning from figaro

PR #1004, opened 2017-07-20
Clean up deprecation warnings, ref #1001

awead:
This makes a number of changes to remove deprecation warnings and other warnings when running RSpec:

* corrects Figaro warnings due to ignored or incorrect variable values
* removes the deprecation warning on FileSet#file_format by using its derived value from Solr
* removes a number of deprecation warnings due to Blacklight::Document preferring hash-like objects
* updates Ldp.contains to RDF::Vocab::LDP.contains
* removes instances of `stub` from RSpec tests

### DIFF
--- a/config/application.yml
+++ b/config/application.yml
@@ -5,21 +5,19 @@
 # Scholarsphere::Config::REQUIREMENTS otherwise, the cap deploy will fail.
 #
 development:
-  TMPDIR: "/tmp"
   ffmpeg_path: "ffmpeg"
   service_instance: "localhost"
   virtual_host: "http://localhost:3000/"
   stats_email: "ScholarSphere Stats <umg-up.its.sas.scholarsphere-email@groups.ucs.psu.edu>"
   google_analytics_id: "test-id"
-  read_only: false
+  read_only: "false"
 test:
-  TMPDIR: "/tmp"
   ffmpeg_path: "ffmpeg-test"
   service_instance: "example-test"
   virtual_host: "http://test.com/"
   stats_email: "Test email"
   google_analytics_id: "test-id"
-  read_only: false
+  read_only: "false"
 production:
   TMPDIR: "/tmp"
   ffmpeg_path: "ffmpeg-test"
@@ -28,4 +26,4 @@ production:
   stats_email: "Test email"
   google_analytics_id: "test-id"
   derivatives_path: "path"
-  read_only: false
+  read_only: "false"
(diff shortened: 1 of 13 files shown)


config/application.yml
```diff
@@ -5,21 +5,19 @@
 # Scholarsphere::Config::REQUIREMENTS otherwise, the cap deploy will fail.
 #
 development:
-  TMPDIR: "/tmp"
```
carolyncole, line 8: @awead what is the reason for removing this?

spec/factories/file_sets.rb
```diff
@@ -43,7 +43,7 @@
 
     trait :with_file_size do
       after(:build) do |fs|
-        fs.stub(:file_size).and_return '1234'
+        allow(fs).to receive(:file_size).and_return("1234")
```
carolyncole, line 46: @awead should still be single quote instead of double quote

spec/presenters/work_show_presenter_spec.rb
```diff
@@ -20,7 +20,7 @@
     end
 
     context "with two files in the work" do
-      let(:solr_doc)    { SolrDocument.new(work.to_solr).merge!('member_ids_ssim' => ["thing1", "thing2"]) }
+      let(:solr_doc)    { SolrDocument.new(work.to_solr).to_h.merge!('member_ids_ssim' => ["thing1", "thing2"]) }
```
carolyncole, line 23: @awead I thought we were changing to respond with a hash.  Why is the to_h needed?  Maybe this should be replaced with a name solr_hash of the change is needed

carolyncole (changes requested):

config/application.yml
awead: Locally, `TMPDIR` is already defined and Figaro issues a warning. Removing this avoids the warning. Notice that the production section still has it because it's on our servers that this variable may not be present.

spec/factories/file_sets.rb
awead: No, it's really a style point here.

spec/presenters/work_show_presenter_spec.rb
awead: There were two deprecation warnings related to Blacklight::Document. One of them was for the `to_hash` change, and the other was for this change. This warning specifically said to do this.

spec/factories/file_sets.rb
awead: ok, re-read this. I'm not intending to be argumentative, I just thought this was a question and not a request. This is only occurrence of a single quote in the file. It's double everywhere else. Could we move the discussion of single vs. double to our style guide and look at using Rubocop to enforce this?

PR #1005, opened 2017-07-20
Limit selection of representative file sets, ref #892

awead:
Selection of representative file sets should be limited to those whose visibility matches that of the work.

This alters some of the decisions that were made in 7fce7edfd211e9fa1d64f21972ec9782afa867ce. Instead of allowing less-visible representatives, only equally visible file sets will be allowed for representatives or thumbnails.

### DIFF
--- a/app/forms/curation_concerns/generic_work_form.rb
+++ b/app/forms/curation_concerns/generic_work_form.rb
@@ -26,14 +26,10 @@ module CurationConcerns
     end
 
     def select_files
-      Hash[file_presenters.map { |file| [name_for_select_file(file), file.id] }]
-    end
-
-    private
-
-      def name_for_select_file(file)
-        return file.to_s unless model.visibility == "open" && file.solr_document.visibility == "authenticated"
-        [file, I18n.t("scholarsphere.select_file_restriction")].join(" ").to_s
+      available_files = file_presenters.select do |file|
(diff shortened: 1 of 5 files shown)


app/forms/curation_concerns/generic_work_form.rb
```diff
@@ -26,14 +26,10 @@ def target_selector
...
-      Hash[file_presenters.map { |file| [name_for_select_file(file), file.id] }]
-    end
-
-    private
-
-      def name_for_select_file(file)
-        return file.to_s unless model.visibility == "open" && file.solr_document.visibility == "authenticated"
-        [file, I18n.t("scholarsphere.select_file_restriction")].join(" ").to_s
+      available_files = file_presenters.select do |file|
+        model.visibility == file.solr_document.visibility
```
carolyncole, line 30: @awead does it need to match exactly or would public be ok if the work was PSU only?

carolyncole (changes requested):

app/forms/curation_concerns/generic_work_form.rb
awead: @cam156 yeah, that crossed my mind too, but why? We don't surface file sets, so having a public file set without a public work doesn't really do much. Although, there's no problem with doing it. I was being lazy and avoiding additional logic 😄

PR #1006, opened 2017-07-20
Prefer double-quotes

awead:
Using single quotes is permitted if their semantics are needed. In all other cases, double quotes are used.

Thoughts? @cam156 @mtribone

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -73,7 +73,10 @@ Style/RegexpLiteral:
   Enabled: false
 
 Style/StringLiterals:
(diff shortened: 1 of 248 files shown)


mtribone:
I generally use double quotes for strings, me thinks, but I'll look at the files to get an idea of the style corrections.

carolyncole:
@awead unfortunately I merged an earlier PR which now causes conflicts with this.

awead:
@cam156 no worries. it's easy to redo!

PR #1007, opened 2017-07-21
Single Value Title and Subtitle

mtribone:
Updates generic work form spec to check each paramater to make sure that they do not contain spaces. Removes empty array from generic work field since the title is displayed as a single value string.

Adds subtitle to the model, new/edit form, show page, property to be displayed, field configurator. Removes subtitle from attribute rows, adds subtitle to the show page under the title, and creates new subtitle partial.

Fixes the catalog controller and field configurator tests for the addition of subtitle. Reworks primary terms method to include subtitle unerneath title in the required metadata section. Adds headings to sufia.en.yml file for the new/edit work form. Updates tests to check for the presence of subtitle.

Updates tests to check for the presence of subtitle.

### DIFF
--- a/app/services/field_configurator.rb
+++ b/app/services/field_configurator.rb
@@ -19,7 +19,8 @@ class FieldConfigurator
              date_created: FieldConfig.new("Date Created"),
              rights: FieldConfig.new("Rights"),
              identifier: FieldConfig.new("Identifier"),
-             description: FieldConfig.new("Description"))
+             description: FieldConfig.new("Description"),
+             subtitle: FieldConfig.new("Subtitle"))
   end
 
   # Defines what fields are shown in the facets on the search index view
--- a/app/views/curation_concerns/base/_work_subtitle.erb
+++ b/app/views/curation_concerns/base/_work_subtitle.erb
@@ -1,0 +1,3 @@
+<% if presenter.subtitle.present? %>
+    <h2><%= presenter.subtitle.first %></h2>
+<% end %>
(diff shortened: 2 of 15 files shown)


mtribone:
screen shot 2017-07-21 at 10 29 02 am

mtribone:
screen shot 2017-07-21 at 10 28 42 am

mtribone:
screen shot 2017-07-21 at 10 30 41 am

app/services/field_configurator.rb
```diff
@@ -19,7 +19,8 @@ def self.show_fields
              date_created: FieldConfig.new("Date Created"),
              rights: FieldConfig.new("Rights"),
              identifier: FieldConfig.new("Identifier"),
-             description: FieldConfig.new("Description"))
+             description: FieldConfig.new("Description"),
+             subtitle: FieldConfig.new(label: "Subtitle", index_solr_type: :stored_searchable))
```
awead: I think `stored_searchable` is the default and can be removed here.

app/views/curation_concerns/base/_work_subtitle.erb
```diff
@@ -0,0 +1,3 @@
+<% if @presenter.subtitle.present? %>
```
awead: Because you've passed in `@presenter` as the local variable `presenter` (see below) you ought to use the local one instead.

awead (changes requested):

app/views/curation_concerns/base/_work_subtitle.erb
awead: I meant "see below" as in this: https://github.com/psu-stewardship/scholarsphere/pull/1007/files#diff-e8df3522061376ee39e0afe98507f5e0R21

app/services/field_configurator.rb
mtribone: Then I can probably remove `label` too.

awead: oh yeah! that too...

awead (approved):

Issue #974, opened 2017-07-06
Rename failed FileSet when URL expires


carolyncole:
When we get an expired url from browse everything we should set the FileSet title to indicate the failure.

'Timeout Error on [filename]. Delete and reupload'

carolyncole:
The error should be I18N so we can easily update the user.

carolyncole:
Add the FileSet rename to here: https://github.com/psu-stewardship/scholarsphere/blob/develop/app/jobs/import_url_job.rb#L28-L32

awead:
@cam156 This was merged in https://github.com/psu-stewardship/scholarsphere/pull/1013 maybe it was moved back to in progress during a retrospective, but I'm not seeing any notes as to why.

Issue #969, opened 2017-06-28
File name not showing properly


olendorf:
For the work https://scholarsphere.psu.edu/concern/generic_works/9kd17cs76p one or more files don't seem to display properly. 

Also looks like file characterization failed. 

Looks like /concern/parent/9kd17cs76p/file_sets/25425k992r in the file list

and is blank on the fileset show page

https://scholarsphere.psu.edu/concern/parent/9kd17cs76p/file_sets/25425k992r

awead:
ImportUrlJob failed with Resque::DirtyExit: Child process received unhandled signal. They must have been adding files via Browse Everything. They will need to delete the file set and try adding it again. There's no way, I think, to retry this because the url from BE will have expired.

The fix for this should be a change to the ImportUrlJob so that failures do not cause the file set to be added to the work, and instead there's a user notification informing them that the file failed to be added and try again.

awead:
Here's another work with similar problems:

https://scholarsphere.psu.edu/concern/generic_works/4mw22v483z

All these files had some kind of error in Resque, even the ones that were successfully added to the work. I believe this is due to the size of the files. Errors ranged from dirty exits, no memory errors and timeouts from hydra-derivatives.

awead:
Of the file sets that we not successfully added to the work, the commonality was either dirty exit or no memory. The no memory errors (~3) all have stack traces that look like:

``` bash
/opt/heracles/deploy/scholarsphere/shared/bundle/ruby/2.3.0/gems/httparty-0.14.0/lib/httparty/request.rb:127:in `join'
/opt/heracles/deploy/scholarsphere/shared/bundle/ruby/2.3.0/gems/httparty-0.14.0/lib/httparty/request.rb:127:in `block in perform'
/opt/heracles/deploy/.rbenv/versions/2.3.3/lib/ruby/2.3.0/net/http.rb:1446:in `block in transport_request'
/opt/heracles/deploy/.rbenv/versions/2.3.3/lib/ruby/2.3.0/net/http/response.rb:163:in `reading_body'
/opt/heracles/deploy/.rbenv/versions/2.3.3/lib/ruby/2.3.0/net/http.rb:1445:in `transport_request'
/opt/heracles/deploy/.rbenv/versions/2.3.3/lib/ruby/2.3.0/net/http.rb:1407:in `request'
[...]

PR #1013, opened 2017-07-27
Rename file sets during upload errors, ref #974 & #969

awead:
ImportUrlJob will fail if the file's url has expired, or if there is some other kind of error during its retrieval, such as a timeout or other disruption.

Instead of letting the job fail and having Resque capture the error, we can have the job capture the error and rename the file set to the kind of error.

The benefit is that it surfaces the error to the user so that they can take action: delete the file set and try again. The drawback is that it's still creating the resource in Fedora. Ideally, perhaps, we ought to capture these errors even before the file set is created and attached to the work. However, that would require deeper changes to the stack.

### DIFF
--- a/app/jobs/import_url_job.rb
+++ b/app/jobs/import_url_job.rb
@@ -1,4 +1,13 @@
 # frozen_string_literal: true
+# Overrides the CurationConcerns job to accept the remote file's original name and handle failures.
+#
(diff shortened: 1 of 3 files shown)


awead:
I've updated this based on suggestions from @olendorf and used i18n for the file set titles and error messages.

PR #1021, opened 2017-08-17
Updating rubocop for security error

carolyncole:
Fixed some errors that were auto-correctable.
Added a TODO for the rest of the new errors.


lib/tasks/scholarsphere/users.rake
```diff
@@ -23,11 +24,11 @@ namespace :scholarsphere do
...
 
-    desc "Report users quota in SS"
+    desc 'Report users quota in SS'
     task quota_report: :environment do
-      caution_sz = 3_000_000_000   # 3GB
-      warning_sz = 5_000_000_000   # 5GB
-      problem_sz = 10_000_000_000  # 10GB
+      # caution_sz = 3_000_000_000   # 3GB
+      # warning_sz = 5_000_000_000   # 5GB
+      # problem_sz = 10_000_000_000  # 10GB
```
awead, line 31: this doesn't look right.

awead (changes requested):
Looks good... lots of todos will make good refactoring fodder. I found only one suspect change, which I commented on.

lib/tasks/scholarsphere/users.rake
carolyncole: @awead rubocop was complaining about unnecessary assignments.  I just commented them out becuase they were not used, but I thought the concept might be useful to keep around.

awead (approved):

Issue #961, opened 2017-06-20
Fedora Object for the User


carolyncole:
An Object in Fedora that has the following fields:
    Given Name ( exists as First Name)
    Surname (exists as Last Name)
    email (optional)
    ORCID (optional)
    login (optional)

Use Friend of Friend ontology model for our predicates.

child of #764

carolyncole:
We may want to pull in other predicates from Friend of Friend like: department, title.

carolyncole:
Linked open data example from ORCID: http://orcid.org/0000-0003-0837-2362

carolyncole:
We will want to index all fields in solr  for searching

carolyncole:
1. Look at FOF for predicates
1. Create Object Code
1. Create Object Index

carolyncole:
Login is the psu id for the person.  We need a rdf predicate for this.

carolyncole:
Changing from first name, middle name to given name and family name.

val99erie:
We submitted a PR that converts creator to a Person model instead of a String, which fulfills part of this story.
https://github.com/psu-stewardship/scholarsphere/pull/1026/commits

In order to complete this story, these fields still need to be added to the Person model and creator partials:
- [ ] ORCID ID
- [ ] PSU login

We don't need to add the field(s) for the alias list as part of this story.  Alias list should be added as part of stories #991 and #992.

awead:
Put Orcid id in the model but not in the UI. The nested form in the work edit page should include PSU id and email.

PR #1026, opened 2017-09-07
Use a Person model for a work's creator instead of a String.  Part of story #961.

val99erie:
This PR includes these changes:

* For the creator of a work (or collection), we changed the creator to be a Person model instead of a String.  The Person model is just a standard `ActiveFedora::Base` record.

* The Person model has properties for `first_name` and `last_name`.  In the future, there are plans to add properties for ORCID ID, Penn State login, and a list of aliases for different display names.

* The work and collection models used to `include ::CurationConcerns::BasicMetadata`, but now they `include ::BasicMetadata` instead.  Since the `creator` property was defined in `::CurationConcerns::BasicMetadata`, we can no longer include that file because we need to define that property differently.  As far as I know, there isn't a way to undefine a property on a model.  So, we copied the file over from curation_concerns, and removed the old `creator` property so that we could re-define it.

* On the new/edit form for a work (or collection), adding/removing a creator will now add/remove a set of fields for the associated Person record.  The user has the option to try to find an existing Person record (and the auto-suggest will fill in the creator fields), or they can enter new data (and a new Person record will be created when the work is saved).

* For the views, presenters, and solr indexing:  The solr index keys have changed.  The keys are now called `creator_name_*` because the old keys, `creator_*` are now used to store the relationship to the Person record.  So now `creator_name_ssim` will index the Person's first and last name, and `creator_sim` will index the URI to the Person record.

* There is a PersonsController that the javascript uses to query Person names for the type-ahead auto-suggest.

* There was some import/export code, mostly under `lib/scholarsphere/import` and some associated files under `spec/fixtures`.  We decided not to spend the effort to update that code for the new Person records because that importer isn't being used anymore.  So, we deleted all that code.

Additional Notes:

* Note that the User model still exists and is still used for authentication.  The User model has not been changed in any way.  Currently there is no association between a logged in user and a Person record.  If there is a need for an association in the future, you can add a relationship or match the records based on Penn State login.

* Currently there is no way for a user to edit an existing Person record.


little9:
This uses the minimal templating system <a href="https://github.com/janl/mustache.js#usage">mustache.js</a> for rendering the repeatable Creator field. We found using a Creator object in JS that shares many of the same fields as the ActiveFedora model and then rendering that was more reliable than copying the existing node in the DOM and appending it. 

<a href="https://github.com/psu-stewardship/scholarsphere/blob/7cf46636deec9d76a24343f0fad4e15c971a5dcb/app/views/records/edit_fields/_creator_template.html.erb">This Rails view</a> contains the mustache template and is not visible on the page. Because it's a Rails view, it's still an ERB template so you can render server-side data in the template. Here's an example: 
```
    <label for="<%= @form.model_class_name %>{{index}}_first_name">Given Name</label>
```

Rails will render `@form.model_class_name` on the page and then the JS will render `{{index}}` .

It also uses the <a href="https://github.com/corejavascript/typeahead.js">typeahead</a> library for autocomplete functionality.

Issue #1011, opened 2017-07-26
Migrate Multiple Titles to Title and Subtitle


carolyncole:
refs #369

We need to migrate data that exists in the current system with multiple titles to the new single title and subtitle with separator.

carolyncole:
Write a rake task to be run once when the single title and subtitle is deployed
Check the object and see if there is more than one title and put that into the subtitle field

carolyncole:
We should have some way to verify the migration occurred correctly.  Possibly a solr query or something?

carolyncole:
@olendorf we have 17 works with more than two titles.  I just want to double check that it is ok to select the first item as the title, and the concatenate the rest with semi colons for the sub title

http://scholarsphere.psu.edu/concern/generic_works/k0698751r
title: schlres_techreport.Wave1.doc
sub title: Prosper_SRI_Wave3_Techreport_V2.pdf; Q#
http://scholarsphere.psu.edu/concern/generic_works/h415p953x
title: Graduate Studies Journal of Organizational Dynamics
sub title: The Environmental Sustainability of Paper; Environmental Sustainability of Paper (1).pdf
http://scholarsphere.psu.edu/concern/generic_works/pg15bd892
title: Reconstructing Woman: From Fiction to Reality in the Nineteenth-Century French Novel
sub title: FM 1..8; Cover
http://scholarsphere.psu.edu/concern/generic_works/4x51hj01t
title: (2002) The Dependence of Lorentz Boost Generators on the Presence and Nature of Interactions
sub title: Preliminary version: comments welcomed.; The Dependence of Lorentz Boost Generators on the Presence and
http://scholarsphere.psu.edu/concern/generic_works/sf268j40c
title: Studies of Anomalies from Part 1
sub title: Trace Metals in Stream Sediment of Southeastern Pennsylvania; The Allenton-Bethlehem Region and the Pennsylvania-Maryland Border Region; Bulletin 86; Multiple Regression Analysis of Trace Metal Data
http://scholarsphere.psu.edu/concern/generic_works/8623j124v
title: Fuel Technology - Curriculum and Career
sub title: Circular 14; Evaluation of Present-Day Mining Instruction
http://scholarsphere.psu.edu/concern/generic_works/hq37vn56b
title: Prosper_SRI_Wave4_Techreport_V1.pdf
sub title: Q#; schlres_techreport.Wave1.doc
http://scholarsphere.psu.edu/concern/generic_works/sf268m02c
[...]

PR #1029, opened 2017-09-12
Adding rake task to migrate titles

carolyncole:
I believe this code is correct, but I double checked what to do with many subtitles on #1011 and want to see the response before this is merged.

That said I took the tact of throw away code here, since it will be run once and then can be tossed.   That means I just wrote a two rake tasks: dry_run & run.  Dry_run is the test to see if it will do the expected thing and also the test to be certain the task ran correctly.  Run is the actual execution.


carolyncole:
Checked with @olendorf and the approach for multiple subtitles should be good to go.

PR #1031, opened 2017-09-13
Change first and last name methods

awead:
First and last name methods did not match the application's view of the terms. Methods and names for "first" name were changed to "given" name; "last" name was changed to "sur" name.


awead:
@cam156 I split this out into its own PR to try to keep things small. It should only be renaming the name fields and nothing else.

carolyncole (approved):

PR #1033, opened 2017-09-13
Adding Coveralls

awead:
We have not been tracking test coverage. This allows us to see if any
changes, pull requests, or commits increase or decrease our overall test
coverage.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -45,6 +45,7 @@ gem 'whenever'
 gem 'yaml_db'
 
 group :development, :test do
+  gem 'coveralls', require: false
   gem 'fcrepo_wrapper'
   gem 'faker'
   gem 'rspec'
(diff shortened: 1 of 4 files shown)


carolyncole:
@awead The changes look great.  I wonder, does the coveralls numbers get affected by how we have split the testing?  Additionally should we put it in only one build?  Maybe along with rubocop?

awead:
@cam156 that's a good question. If we did move to one build, it would slow things down again. I am now noticing that the coverage appears twice, maybe because of the two builds?

carolyncole:
@awead for grins and giggles I tried using the wear_merged! in this branch: https://github.com/psu-stewardship/scholarsphere/blob/coveralls2/spec/rails_helper.rb#L9
It gave 82%.

PR #1036, opened 2017-09-15
Refactor UserStatsImporter

awead:
From our Friday refactoring session, we extracted the various tally
methods into a new class and removed the inheritance from
Sufia::UserStatImporter.

### DIFF
--- a/app/services/user_stats_importer.rb
+++ b/app/services/user_stats_importer.rb
@@ -1,12 +1,9 @@
 # frozen_string_literal: true
 
-class UserStatsImporter < Sufia::UserStatImporter
-  attr_reader :start_date, :end_date, :view_report, :download_report
+class UserStatsImporter
+  attr_reader :view_report, :download_report
 
-  def initialize(start_date, end_date = 1.day.ago, options = {})
-    super(options)
-    @start_date = start_date
-    @end_date = end_date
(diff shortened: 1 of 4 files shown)


app/services/user_stats_importer.rb
```diff
@@ -118,4 +79,17 @@ def create_or_update_object_stat(event, user_db_id, object, stat_class)
                      end
       updated_stat.save
     end
+
+    def create_or_update_user_stats(stats, user)
```
carolyncole: Should we add a comment that this code was copied from Sufia::UserStat?

carolyncole (changes requested):
@awead At some point we will need to add tests for the new class.  Did you want to add those now or should we add a ticket?  It should be a matter of taking some of the test from UserStats and putting them against UserStatisticReport.

carolyncole:
@awead I'm not certain if this is an accurate code coverage drop.  Are you happy with the coverage on this PR?

awead:
@cam156 yeah, I wasn't either. I'm happy if you are!

carolyncole:
@awead if I am reading code coverage right it looks like import is not being tested
https://coveralls.io/builds/13847428/source?filename=app%2Fservices%2Fuser_stats_importer.rb#L12

Issue #991, opened 2017-07-14
Keep track of aliases


carolyncole:
When a user enters a different display from the default from LDAP or ORCID we should attach that as an Alias to the user Dependent on #989 & #990

carolyncole:
The aliases would be generated by a user changing the display name on the work form or the collection form.

carolyncole:
Add  (display name) alias to the form for the person

carolyncole:
Is the alias a property of the work or an additional object linked to the person.

As person has many Aliases, String or Object, we are leaning towards Objects.

awead:
After discussion today, the prevailing thought is that  Alias will be a property of Work, and Person will be a property of Alias:

``` bash
[Work](1) <--> (n)[Alias](n) <--> (1)[Person]
```

The work form can remain the same, with an added display name field that causes a new alias to be created or an existing one to be re-used. The net result of this would be:

1. A work's creators are all Alias objects
2. A person must have at least one alias in Fedora, even if it is to display only their first and last name with no changes from their person record.
3. When the same display name is used in a different work, the same alias should be used. Ideally, we should not create multiple alias resources with the same display name.

## Impacts

Work presently links to Person resources, so this would need to be redefined in the `HasCreators` module, and changed to `has_and_belongs_to_many :aliases` for example. The alias would then have a display name property and link to Person:

``` ruby
class Alias < ActiveFedora::Base
  property :person, class_name: 'Person', predicate: ::RDF::Vocab::FOAF.[TBD]
  
  property :display_name, predicate: ::RDF::Vocab::FOAF.[TBD], multiple: false do |index|
    index.as :stored_searchable, :symbol
  end
end
```

PR #1039, opened 2017-09-19
Allow ordering of alias objects in place of persons, ref #991, #949

awead:
This makes two main changes: 1) Creates an Alias resource that stands between a work and its creator; and 2) orders aliases using a membership predicate.

Previously, a work would contain one or more Person objects as creators. In order to allow for different names for the same canonical person, an Alias class holds a display name and a reference to a single canonical user. The `AliasManagementService` is responsible for finding or creating persons and aliases to attach to new or existing works.

Forms for works can now accept either new creators or edit existing ones. When new creators are added, both aliases and persons are created as needed, via the `AliasManagementService`, or if display names are changed, new aliases are added to the existing person, also via the `AliasManagementService`.

### DIFF
--- a/app/assets/javascripts/scholarsphere/creator/creator.js
+++ b/app/assets/javascripts/scholarsphere/creator/creator.js
@@ -4,7 +4,8 @@ ScholarSphere.creator = {
    */
   id: '',
   firstName: '',
-  lastName: '', displayName: '',
+  lastName: '',
+  displayName: '',
   index: '',
   readonly: ''
 }
--- a/app/assets/javascripts/scholarsphere/creator/creator.js
+++ b/app/assets/javascripts/scholarsphere/creator/creator.js
@@ -4,7 +4,7 @@ ScholarSphere.creator = {
    */
   id: '',
   firstName: '',
-  lastName: '',
+  lastName: '', displayName: '',
   index: '',
   readonly: ''
 }
--- a/app/services/generic_work_to_share_json_service.rb
+++ b/app/services/generic_work_to_share_json_service.rb
@@ -22,7 +22,7 @@ class GenericWorkToShareJSONService
 
     def add_contributors_to_document
       work.creators.each do |creator|
-        creator_name = [creator.sur_name, creator.given_name].compact.join(', ')
+        creator_name = [creator.person.sur_name, creator.person.given_name].compact.join(', ')
         document.add_contributor(name: creator_name, email: email_for_name(creator_name))
       end
     end
--- a/spec/features/dashboard/dashboard_collections_spec.rb
+++ b/spec/features/dashboard/dashboard_collections_spec.rb
@@ -6,11 +6,9 @@ include Selectors::Dashboard
 
 describe 'Dashboard Collections:', type: :feature do
   let!(:jill_collection) { create(:collection, title: ["Jill's Collection"], depositor: jill.login) }
-  let!(:collection)      { create(:collection, depositor: current_user.login) }
+  let!(:collection)      { create(:collection, creators: [creator], depositor: current_user.login) }
 
-  let(:creator) { collection.creators.first }
-  let(:creator_name) { [creator.given_name, creator.sur_name].join(' ') }
-
+  let(:creator)      { create(:alias, :with_person) }
   let(:current_user) { create(:user) }
   let(:jill)         { create(:jill) }
 
@@ -33,14 +31,14 @@ describe 'Dashboard Collections:', type: :feature do
 
   specify 'toggle displays additional information' do
     first('span.glyphicon-chevron-right').click
-    expect(page).to have_content(creator_name)
+    expect(page).to have_content(creator.display_name)
     expect(page).to have_content(collection.depositor)
     expect(page).to have_content 'Edit Access'
     expect(page).to have_content(current_user)
   end
 
   specify 'additional information is hidden' do
-    expect(page).not_to have_content(creator_name)
+    expect(page).not_to have_content(creator.display_name)
     expect(page).not_to have_content(collection.depositor)
     expect(page).not_to have_content 'Edit Access'
     expect(page).not_to have_content(current_user)
@@ -67,7 +65,7 @@ describe 'Dashboard Collections:', type: :feature do
         click_link('Object Type')
         expect(page).to have_content('Collection (1)')
         click_link('Creator')
-        expect(page).to have_content(creator_name)
+        expect(page).to have_content(creator.display_name)
       end
     end
   end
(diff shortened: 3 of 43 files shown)


app/assets/javascripts/scholarsphere/creator/creator.js
```diff
@@ -4,7 +4,7 @@ ScholarSphere.creator = {
    */
   id: '',
   firstName: '',
-  lastName: '',
+  lastName: '', displayName: '',
```
carolyncole: A Style comment.  Since every item is on it's own line it would be better to add the displayName on a separate line.

app/services/generic_work_to_share_json_service.rb
```diff
@@ -22,7 +22,7 @@ def json
 
     def add_contributors_to_document
       work.creators.each do |creator|
-        creator_name = [creator.sur_name, creator.given_name].compact.join(', ')
+        creator_name = [creator.person.sur_name, creator.person.given_name].compact.join(', ')
```
carolyncole, line 25: Should this really just be display_name?

spec/features/dashboard/dashboard_collections_spec.rb
```diff
@@ -33,14 +31,14 @@
 
   specify 'toggle displays additional information' do
     first('span.glyphicon-chevron-right').click
-    expect(page).to have_content(creator_name)
+    expect(page).to have_content(collection.creator.first.display_name)
```
carolyncole: Could this just be `expect(page).to have_content(creator.display_name)`

spec/features/dashboard/dashboard_works_spec.rb
```diff
@@ -5,25 +5,28 @@
...
+
+  let(:creator) { build(:alias, display_name: work1_creator_name,
+                                person: Person.new(given_name: 'Creator1', sur_name: 'Jones')) }
 
   let!(:work1) do
     create(:public_work, :with_complete_metadata,
+           id: 'dashboard-works-work1',
            depositor: current_user.login,
            title: ['little_file.txt'],
-           creators: [create(:creator)],
```
carolyncole, line 15: @awead I think the reason for using a different creator was to make sure the query was filtering.  When you put in the same creator we have no idea if we are getting all or just the records we want.

spec/forms/generic_work_form_spec.rb
```diff
@@ -8,26 +8,26 @@
...
-        expect(form.creators.count).to eq 1
-        current_creator = form.creators.first
-        expect(current_creator.given_name).to eq 'Test A'
-        expect(current_creator.sur_name).to eq 'User'
-      end
+  describe '#creators' do
+    subject { form.creators.first }
+
+    context 'without any existing creators' do
+      its(:display_name) { is_expected.to eq('Test A User') }
```
carolyncole, line 15: Doesn't the form still show the given_name and sur_name?

spec/models/generic_work_spec.rb
```diff
@@ -12,75 +12,85 @@
...
       end
     end
 
     context 'building a creator' do
       let(:work) { build(:work) }
 
-      it 'creates a Person record' do
-        work.creators.build(given_name: 'Frodo')
+      it 'creates an Alias record' do
+        work.creators.build(display_name: 'Frodo')
```
carolyncole: Can we build with a first and last name?

carolyncole (changes requested):
Overall looks very good.  Added a few inline comments.  My big overall question is it seems like you can not create a Person/Alias by filling in the user Given and Surname.  The goal is not to need to parse a string every time, but allow the user to give us that information in parts.  I can not tell, but it seems like this PR is removing that ability.

app/assets/javascripts/scholarsphere/creator/creator.js
mtribone: Certainly. My editor kept indenting it 2 spaces and was annoying me, so I just kept it on the same line.  :-)

app/services/generic_work_to_share_json_service.rb
awead: @cam156 yes, I thought the same too. I was thinking that the we ought to link the SHARE resource off the canonical name of the person, as opposed to the alias, otherwise we might have different users in SHARE that are the same person.

awead:
@cam156 We could make `display_name` optional in the form, which would mean that if it is blank, `AliasManagementService` returns the first alias it finds for the person. `given_name` and `sur_name` would have to be required because that's what we use to find the person--actually, this is what the search form is doing for us, so no problem there.

If a display name is provided, then it would create a new alias if that display name did not already exist.

Is this what you were thinking of?

spec/models/generic_work_spec.rb
awead: @cam156 no, because `build` is expecting an `Alias` which doesn't have those attributes, but we could use `person`, I think, like: `work.creators.build(display_name: 'Frodo', person: Person.new(given_name: 'Frodo', sur_name: 'Baggins'))` but the model doesn't care about the person at this point. The presence or absence of a person is enforced at the `creators=` method via `AliasManagementService`. But, there's no harm in doing the above... it makes it more complete, I guess.

spec/forms/generic_work_form_spec.rb
awead: It shows all three: display, given, and sur. Is that what's supposed to happen?

app/services/generic_work_to_share_json_service.rb
carolyncole: @awead makes sense. There is likely more work in the share module so let's just leave it alone for now.

spec/forms/generic_work_form_spec.rb
carolyncole: @awead I was commenting on the fact that the test only tests display name now.  Seemed like the form could no longer have the other two

carolyncole:
@awead I just got a bit worried becuase all the tests were just checking display name.  I think display name being required is ok, as long as we default it to `"#{surname}, #{given_name}"`

carolyncole:
@awead that said the defaulting does not need to be part of this PR.  Just something we should do to help out the users.

spec/features/dashboard/dashboard_works_spec.rb
awead: @cam56 I put in a different creator name for `work1` which will differ from the creator name used in the facet count check.

spec/forms/generic_work_form_spec.rb
awead: @cam156 I've expanded the tests to include checks for given and surnames.

awead:
@cam156 presently, `display_name` is defaulting to whatever the display name for the record in `User` is, which is what's just coming out of LDAP. For me, this is _ADAM GARNER WEAD_, in all caps and not, for example, _Wead, Adam Garner_. Do we want to change that? The other issue with this is what if a user prefers a particular display name? For example, I always want to be _Dr. Adam Wead_ and I'd have to change that each time I deposit unless I set a preferred alias or have the form look for whatever my current alias is... That would be a good follow-up PR.

carolyncole (approved):

PR #1040, opened 2017-09-19
Collection and Work migration

carolyncole:
I putting this out for comment early.  The work is not complete, but I want to be certain I have broken things up in a logical fashion.  (not just logical to me)

The CreatorMigrator is the main class that calls all the rest to migrate the creators for both Collections and Works.

The SolrListMigrator and the CreatorList are still only partially implemented since we do not have the Alias PR merged.  The CreatorList will convert all the creator Strings to Aliases using the AliasManagementService once that is included into develop.  The SolrListMigrator will use the new Aliases to convert the string creators currently in the system for each object to pointers to Aliases. 

LocalUserLookup is a helper class to identify creators as local users to link or create them with the PSU login information.

SolrObjectlist, SolrWorkList and SolrCollectionList loads all the object ids from solr and then loads each item from fedora as we iterate through the ids.  The idea is to save on memory since Collection.all or GenericWork.all loads all objects into memory when only one is needed at a time.

### DIFF
--- a/app/models/migration/creator_list.rb
+++ b/app/models/migration/creator_list.rb
@@ -1,0 +1,125 @@
+# frozen_string_literal: true
+
+module Migration
+  class CreatorList
+    attr_reader :uniq_system_creators, :creator_alias_cache
+    def initialize(creator_alias_cache = 'tmp/creator_alias_cache.json')
+      @creator_alias_cache = creator_alias_cache
+      @uniq_system_creators = find_uniq_system_creators
+    end
+
+    def to_alias_hash
+      @agents ||= convert_to_aliases
+    end
+
+    private
+
+      def find_uniq_system_creators
+        creator_facet_resp = ActiveFedora::SolrService.instance.conn.get 'select', params: { rows: 0, 'facet.limit' => 10000, 'facet.field' => 'creator_sim' }
+        creator_facet_resp['facet_counts']['facet_fields']['creator_sim'].select { |creator_or_count| creator_or_count.class == String }.reject { |name_or_url| name_or_url.starts_with?('Http') }
+      end
+
+      def convert_to_aliases
+        aliases = load_cache
+        @uniq_system_creators.map do |creator|
+          next if aliases.key?(creator)
+          aliases[creator] = creator_to_alias(creator)
+        end
+        store_cache(aliases)
+        aliases
+      end
+
+      def load_cache
+        return {} unless File.exist?(creator_alias_cache)
+
+        aliases = {}
+        file = File.open(creator_alias_cache)
+        lines = file.readlines
+        lines.each do |line|
+          creator, alias_id = line.strip.split(cache_separator)
+          aliases[creator] = Alias.find(alias_id)
+        end
+        aliases
+      end
+
+      def store_cache(aliases)
+        file = File.new(creator_alias_cache, 'w')
+        aliases.each_pair do |current_creator, current_alias|
+          file.puts("#{current_creator}#{cache_separator}#{current_alias.id}")
+        end
+        file.close
+        aliases
+      end
+
+      def creator_to_alias(creator)
+        users = LocalUserLookup.find_users(creator)
+        if users.count == 1
+          user = users[0]
+          # validate creator against user
+          # todo Not really sure what to do here
+          # logger.warn("Matching #{creator} with #{user.display_name} #{user.login}")
+          #
+          # user_to_alias(user, creator)
+          if validate_user(creator, user)
+            logger.warn("Matching #{creator} with #{user.display_name} #{user.login}")
+            user_to_alias(user, creator)
+          else
+            logger.warn("No match #{creator} with #{user.display_name} #{user.login} did not validate.  Creating new agent")
+            name_to_alias(creator)
+          end
+        else
+          name_to_alias(creator)
+        end
+      end
+
+      def validate_user(creator, user)
+        # id match is valid
+        return true if creator == user.login
+
+        logger.warn("Matching #{creator} with #{user.display_name} #{user.login}")
+        user_name = parse_name(user.display_name)
+        creator_name = parse_name(creator)
+        user_name.family.downcase!
+        creator_name.family.downcase!
+
+        # family name match is valid
+        return true if user_name.family == creator_name.family
+
+        comparable_name(creator) == comparable_name(user.display_name)
+      end
+
+      def comparable_name(name)
+        name_parts = name.downcase.gsub(',', ' ').squeeze(' ').split(' ')
+        comparable = Hash.new(0)
+        name_parts.each { |v| comparable[v] += 1 }
+        comparable
+      end
+
+      def name_to_alias(full_name)
+        name = parse_name(full_name)
+        AliasManagementService.call(display_name: full_name, given_name: name.given, sur_name: name.family)
+      end
+
+      def parse_name(full_name)
+        name = Namae::Name.parse(full_name)
+        if name.given.blank? || name.family.blank?
+          name.family = full_name
+          name.given = nil
+        end
+        name
+      end
+
+      def user_to_alias(user, creator)
+        agent = Agent.where(psu_id: user.login).first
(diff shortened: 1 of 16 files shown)


app/models/migration/creator_list.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module Migration
+  class CreatorList
+    attr_reader :creators
+    def initialize
+      creator_facet_resp = ActiveFedora::SolrService.instance.conn.get 'select', params: { rows: 0, 'facet.limit' => 10000, 'facet.field' => 'creator_sim' }
+      @creators = creator_facet_resp['facet_counts']['facet_fields']['creator_sim'].select { |creator_or_count| creator_or_count.class == String }
```
awead: I can't tell, but this looks like it's an array of all `creator_sim` values in Solr. Maybe some YARD docs to clear that up?

app/models/migration/creator_list.rb
```diff
@@ -0,0 +1,33 @@
...
+        end
+      end
+
+      def creator_to_alias(creator)
+        users = LocalUserLookup.find_users(creator)
+        if users.count == 1
+          # validate creator against user
+          # find agent by login
+          # add alias to agent
+        end
```
awead: Do we care about any failures or missed creators at this point? I'm not sure if we do either :)

app/models/migration/solr_list_migrator.rb
```diff
@@ -0,0 +1,33 @@
...
+      def migrate_creators(solr_object_list, creator_alias_hash)
+        solr_object_list.each_with_load do |object|
+          update(object, creator_alias_hash)
+        end
+      end
+
+      def update(object, creator_alias_hash)
+        aliases = update_creators(object.creators, creator_alias_hash)
+        # assign alias list here
+        # put the old creators into the creator_name predicate
```
awead: is creator_name on the alias?

awead (commented):
I think this looks fine. I only had a few questions. Couple of general thoughts: is this going to be idempotent? Can we re-run it multiple times if we need to pick up more folks? I'm guess we won't care too much about missed creators because we'll see who's missing at the end and pick them up later.

app/models/migration/creator_list.rb
carolyncole: @awead  I think we do care.  We will need to create one for each alias.

app/models/migration/creator_list.rb
carolyncole: @awead I can rename the variable. How about uniq_system_creators?

app/models/migration/solr_list_migrator.rb
carolyncole: @awead I'm not certain.  Things got moved around a fair amount. Originally there was a list of creator_names (display_names) on the Person.  Now I guess that is part of the Alias in some fashion.

awead: @cam156 yes, that's right. I was just making sure I understood. There's a `display_name` property on the Alias, which I'm assuming is `creator_name` here.

carolyncole:
@awead I was not really thinking about rerunning it, but that is a good thought.  I wonder if we could create the aliases once and then store and load them from file.  That way if the process fails we could restart.

app/models/migration/creator_list.rb
```diff
@@ -0,0 +1,79 @@
...
+          # todo should really be looking up person by id
+          name_to_alias(user.display_name, creator)
+        else
+          name_to_alias(creator, creator)
+        end
+      end
+
+      def name_to_alias(full_name, creator)
+        name = Namae::Name.parse(full_name)
+        name.family = full_name if name.given.blank?
```
awead: shouldn't this be: `name.family ||= full_name` to avoid a missing surname?

awead: or `name.family = full_name if name.family.blank?`

app/models/migration/solr_object_list.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module Migration
+  class SolrObjectList
+    attr_reader :objects, :object_class
+
+    def initialize(object_class)
+      @object_class = object_class
+      solr_resp = ActiveFedora::SolrService.instance.conn.get 'select', params: { rows: 1000, fl: 'id,depositor_tesim', q: '', fq: filter }
+      @objects = solr_resp['response']['docs'].map { |doc| doc['id'] }
```
awead: this can be simplified to:
``` ruby
@object = ActiveFedora::SolrService.query(filter, fl: 'id,depositor_tesim', rows: 1000).map(&:id)
```

app/models/migration/solr_list_migrator.rb
```diff
@@ -0,0 +1,42 @@
...
+    class << self
+      def migrate_creators(solr_object_list, creator_alias_hash)
+        solr_object_list.each_with_load do |object|
+          update(object, creator_alias_hash)
+        end
+      end
+
+      def update(object, creator_alias_hash)
+        return if object.creator_ids.empty?
+        return if !object.creator_ids[0].match(/^[0-9A-F]{8}-[0-9A-F]{4}/i).nil?
```
awead: why are we only concerned with the first id?

app/models/migration/creator_list.rb
```diff
@@ -0,0 +1,79 @@
...
+        person = Person.where(given_name: name.given, sur_name: name.family).first
+        if person.blank?
+          person = Person.create(given_name: name.given, sur_name: name.family)
+        end
+        Alias.create(display_name: creator, person: person)
```
awead: all of the above could be replaced with:
``` ruby
AliasManagementService.call(display_name: creator, given_name: name.given, sur_name: name.family)
```
Which would create the alias and find or create the person, but it's unclear if we want to allow missing surnames at this point. The service would raise an error if surname was missing.

app/models/migration/creator_list.rb
```diff
@@ -0,0 +1,79 @@
...
+          creator, alias_id = line.strip.split('||')
+          aliases[creator] = Alias.find(alias_id)
+        end
+        aliases
+      end
+
+      def store_cache(aliases)
+        file = File.new(creator_alias_cache, 'w')
+        aliases.each_pair do |current_creator, current_alias|
+          file.write("#{current_creator}||#{current_alias.id}\n")
```
awead: `file.puts("#{current_creator || current_alias.id}")` better? The `puts` should put the return.

spec/features/migration_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    work2
+
+    save_collection_to_solr(collection, collection_creator)
+    save_work_to_solr_and_fake_fedora(work, work_creator)
+    allow(collection).to receive(:creator_ids).and_return([collection_creator])
+  end
+
+  after do
+    ActiveFedora::Cleaner.cleanout_solr
+    File.delete(cache_file) if File.exist?(cache_file)
```
awead: `FileUtils.rm_f(cache_file)` does the same and doesn't bark if the file doesn't exist.

spec/models/migration/creator_list_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+    it { is_expected.to eq ['abc for me', 'abc for me too'] }
+  end
+
+  describe '#to_alias_hash' do
+    subject(:creator_list) { described_class.new(cache_name).to_alias_hash }
+
+    let(:user) { create :user, display_name: 'Frog, Kermit The' }
+    let(:work4) { build :work, id: '999zzz' }
+    let(:cache_name) { 'tmp/test_cache.txt' }
+    let(:person1) { Person.new(sur_name: 'blarg institution for the insane') }
```
awead: hmm. is work getting to you?

awead (commented):
I had a couple of questions and some suggestions. I had a hard time following all the private methods in CreatorList, but since this migration code and not likely to be used much later, I'm less inclined to devote cycles to it just to clean it up. If works... it's good.

PR #1056, opened 2017-10-24
Pagination of the filesets for Work Show Page

mtribone:
Updates paginate to use the member presenters and to correct the route for the main app.

Fixes up the paginate partial to display on screen the paginator.

Updates controller to use file_page instead of page because page is being used by something else.

Updates spec test to check for 10 files on page 1 and look for a link for the second page.

### DIFF
--- a/app/views/curation_concerns/base/_paginate.html.erb
+++ b/app/views/curation_concerns/base/_paginate.html.erb
@@ -1,0 +1,11 @@
+<%# Need to pinch the paginator from
+    https://github.com/samvera/curation_concerns/blob/master/app/views/collections/_paginate.html.erb
+    because the fieldsets for works are ordered and not a solr response. Uses the presenter and a
+    new method of file_page to create pages of objects
+%>
+<% if @presenter.member_presenters.total_pages > 1 %>
+    <div class="pager">
+      <%= paginate @presenter.member_presenters, route_set: main_app, param_name: 'file_page',  outer_window: 2, theme: 'blacklight' %>
+      <div class="clearfix"></div>
+    </div><!-- /pager -->
+<% end %>
--- a/spec/features/generic_work/work_with_many_files_spec.rb
+++ b/spec/features/generic_work/work_with_many_files_spec.rb
@@ -42,12 +42,20 @@ describe GenericWork, type: :feature do
       visit('/concern/generic_works/bigwork')
     end
 
-    it 'displays the work page with all the files' do
+    it 'displays the work page with pagination and each page has 10 files' do
       within('dl.attributes') do
         expect(page).to have_selector('dd.total_items', text: '100')
       end
       within('table.related-files') do
-        (1..100).each do |id|
+        (1..10).each do |id|
+          expect(page).to have_link("File #{id}")
+        end
+      end
+      within('div.pager') do
+        click_link('2')
+      end
+      within('table.related-files') do
+        (11..20).each do |id|
           expect(page).to have_link("File #{id}")
         end
       end
--- a/app/controllers/curation_concerns/generic_works_controller.rb
+++ b/app/controllers/curation_concerns/generic_works_controller.rb
@@ -16,6 +16,8 @@ class CurationConcerns::GenericWorksController < ApplicationController
   before_action :delete_from_share, only: [:destroy]
   before_action :redirect_when_uploading, only: [:edit, :update, :destroy]
 
+  before_action :pass_page_to_presenter, only: [:show]
+
   def notify_users_of_permission_changes
     return if @curation_concern.nil?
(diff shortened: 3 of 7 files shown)


mtribone:
screen shot 2017-10-24 at 3 44 46 pm

spec/features/generic_work/work_with_many_files_spec.rb
```diff
@@ -47,10 +47,13 @@
         expect(page).to have_selector('dd.total_items', text: '100')
       end
       within('table.related-files') do
-        (1..100).each do |id|
+        (1..10).each do |id|
           expect(page).to have_link("File #{id}")
         end
       end
+      within('div.pager') do
```
carolyncole: @mtribone We may want to click on the page link and check the next page to see if files 11..20 exists on the second page.
Additionally will there be other page links? 2..10?

carolyncole (commented):
@mtribone overall looks good.  The test only tests a bit of what the old test did.  I added a comment inline.

spec/features/generic_work/work_with_many_files_spec.rb
mtribone: Yes there will be other page links. Should we test each one or is the presence of a second page _with_ records enough?

carolyncole: @mtribone testing that the links exist and that we can get to the second page should be enough.  @awead do you agree?

app/views/curation_concerns/base/_paginate.html.erb
```diff
@@ -0,0 +1,6 @@
+<% if @presenter.member_presenters.total_pages > 1 %>
```
awead, line 6: @mtribone is this a partial you're overriding? Can you put in `<%# comment ... %>` on the first line about why?

spec/features/generic_work/work_with_many_files_spec.rb
awead: @mtribone yeah. Also, checking for the link with `have_link` and then clicking it are the same thing. You could probably just say `click_link`. If it's not there or it can't find, it'll break.

spec/features/generic_work/work_with_many_files_spec.rb
```diff
@@ -42,15 +42,28 @@
...
+        expect(page).to have_link('2')
+        click_link('2')
+      end
+      within('table.related-files') do
+        (2..11).each do |id|
```
awead: Shouldn't this be `(11..20)`? I thought we were on the second page of 10 items out of 100.

awead (commented):
Just added a few questions of my own and responded to @cam156's comments. This looks awesome!

spec/features/generic_work/work_with_many_files_spec.rb
mtribone: I went to art school, so math isn't one of my subjects.

app/views/curation_concerns/base/_paginate.html.erb
mtribone: Okay. I believe that I had to pinch the paginate from Curation Concerns collections and use the presenter because the fieldsets are ordered. Whereas the collection listing of works is a Solr response.
https://github.com/samvera/curation_concerns/blob/master/app/views/collections/_paginate.html.erb

mtribone: And give it a new method of `file_page` for page in the work_show_presenter

mtribone: @awead @cam156 Does this comment make sense to you?

"Need to pinch the paginator from https://github.com/samvera/curation_concerns/blob/master/app/views/collections/_paginate.html.erb
    because the fieldsets for works are ordered and not a solr response. Uses the presenter and a new method of file_page to create pages of ordered objects"

awead: @mtribone great!

spec/features/generic_work/work_with_many_files_spec.rb
awead: @mtribone I went to music school, so I can only count to 4. Don't trust my math either... `(2..11)` may in fact be correct. I would think that the test would fail, though, because you'd be looking at files 11 through 20. Try throwing in a `save_and_open_page` and just verify that you're seeing what you're supposed to be seeing.

mtribone: @awead I guess the files were still within the range even though file 11 wasn't present. Save and open page does show me files 11-20.

awead: @mtribone ok. I guess just update the test to check that range and we should be good.

PR #1059, opened 2017-10-25
Allow missing given names in AliasManagementService

awead:
If no given name is provided, a new record will be created using only
the surname. Searching for a nil given name returns only records without
that given name present, avoiding the issue of two existing persons with
the same surname, but one with a missing given name.

### DIFF
--- a/spec/services/alias_management_service_spec.rb
+++ b/spec/services/alias_management_service_spec.rb
@@ -4,9 +4,21 @@ require 'rails_helper'
 
 describe AliasManagementService do
   let(:service) { described_class.call(attributes) }
-  let(:person) { create(:person, given_name: 'Johnny', sur_name: 'Depp') }
   let(:missing_person) { I18n.t('scholarsphere.aliases.person_error') }
   let(:missing_parameter) { I18n.t('scholarsphere.aliases.parameter_error') }
+  let(:johnny_depp) { Person.where(given_name: 'Johnny', sur_name: 'Depp').first }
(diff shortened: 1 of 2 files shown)


spec/services/alias_management_service_spec.rb
```diff
@@ -76,22 +76,35 @@
   end
 
   context 'when the person exists but the alias does not' do
-    before { Alias.destroy_all }
+    let(:new_alias) { Alias.where(display_name: 'Capt. Jack Sparrow').first }
 
-    context 'with all the required parameters' do
+    before do
+      Alias.destroy_all
+      if Person.where(given_name: 'Johnny', sur_name: 'Depp').empty?
```
carolyncole: Why is this needed?  Couldn't you just reference you let?  `person`

carolyncole (changes requested):
Other than the strance create in the before it is all looking good.

spec/services/alias_management_service_spec.rb
awead: @cam156 this is a "find or create" scenario. `let` was creating multiple Johnny Depps. I wanted to ensure there was only one.

carolyncole: @awead you could set the id in the let statement

awead: @cam156 possibly. "Let" me give that a try...

awead:
@cam156 I pushed up a change. Providing the id in the create method wasn't working, it was just making a new object. I reorganized the create/update process so that there will always be one of each record type.

PR #1061, opened 2017-10-26
Additional properties to Person, ref #961

awead:
Adds email, PSU id, and Orcid id to the Person model so we can store
these values when users want to add them.


app/controllers/persons_controller.rb
```diff
@@ -4,8 +4,26 @@ class PersonsController < ApplicationController
...
+
+    def search_results(query)
+      aliases = ActiveFedora::SolrService.query(
+        "display_name_tesim:#{query}*",
+        fq: 'has_model_ssim:Alias',
+        fl: ['id', 'display_name_tesim', 'name_ssim'],
+        qt: 'select',
+        rows: 10
+      )
+      aliases.each do |aliaz|
```
carolyncole: @awead how about agent_alias?  We are never going to know to search for aliaz if we are looking at the code.

app/controllers/persons_controller.rb
```diff
@@ -4,8 +4,26 @@ class PersonsController < ApplicationController
...
+        person = ActiveFedora::SolrService.query(
+          "id:#{aliaz['name_ssim'].first}",
```
carolyncole: @awead how is the name the ID?  That seems strange

app/forms/creator_form.rb
```diff
@@ -30,6 +30,21 @@ def given_name
     end
   end
 
+  def psu_id
+    return unless read_only_name?
```
carolyncole: I'm wondering based on the fact that we are using read_only_name? for things that are not related to name if we should change that to just read_only?

spec/factories/persons.rb
```diff
@@ -5,4 +5,12 @@
     sequence(:given_name) { |n| "First Name #{n}" }
     sequence(:sur_name) { |n| "Last Name #{n}" }
   end
+
+  trait :with_complete_metadata do
+    given_name 'Johnny C.'
+    sur_name   'Lately'
+    email      'newkid@example.com'
+    psu_id     'jcl81'
+    orcid_id   '00123445'
```
carolyncole, line 14: I wonder if we care about validating the orcid.

carolyncole (changes requested):

app/controllers/persons_controller.rb
awead: @cam156 This one got me too. It's generated by ActiveFedora's solr name stuff. Specifically, when you define the relationship using the `name` predicate here:

https://github.com/psu-stewardship/scholarsphere/blob/user-object-961/app/models/alias.rb#L6

I would have expected `person_ssim`, but I guess not.

spec/factories/persons.rb
awead: @cam156 dunno. I imagine we should. We put that in another ticket and validate by some concrete means, like length, etc., and then look at a future option that could query the orchid service directly, like an autocomplete or just a call back that's done at some point.

.rubocop_todo.yml
```diff
@@ -1,6 +1,12 @@
-require: rubocop-rspec
```
carolyncole, line 1: @awead did you intend to remove rubocop-rspec?

awead: hmm. no. I think auto-gen did that.

awead: @cam156 `--auto-gen-config` seems to have done that. I'm not sure why it's showing up now, but the require is in `.rubocop_yml` so it doesn't need to be in the todo file.

awead:
@cam156 I took a look at the coveralls changes, and the missing coverage doesn't seem to be related to any of the changes in this PR.

awead:
@cam156 oohhh-kay.... now it says coverage increased. I rebased it, so maybe that was it.

Issue #998, opened 2017-07-17
Keyword with special characters creates bad search


carolyncole:
When you include special characters in the keyword the search produced by clicking the link on the work is bad
![screen shot 2017-07-17 at 7 04 29 am](https://user-images.githubusercontent.com/1599081/28265546-629f8b60-6abe-11e7-8f23-26bfbc941257.png)

The facet with the keyword seems to work, but however we are creating the link in the work seems bad
Facet Link: `https://scholarsphere-qa.libraries.psu.edu/catalog?f%5Bkeyword_sim%5D%5B%5D=%2755+chet+atkins`
Work Link: `https://scholarsphere-qa.libraries.psu.edu/catalog?f%5Bkeyword_sim%5D%5B%5D=%26%2339%3B55+chet+atkins`

awead:
Fixed. See https://scholarsphere-qa.libraries.psu.edu/concern/generic_works/0p2676t33d

awead:
Doesn't work with featured works.

PR #1064, opened 2017-10-30
Use WorkShowPresenter for recent and featured works, ref #998

awead:
@cam156 @mtribone this alters the views of recent and featured works to make them exactly like the show view. This way, we're using the same renderer code that we use to translate facets. The original problem was identified in #998, but we discovered is also was present in recent and featured works because they weren't using the same renderer process.

The question is, do we need to tweak the layout or display:
featured
recent

### DIFF
--- a/spec/features/featured_work_spec.rb
+++ b/spec/features/featured_work_spec.rb
@@ -5,7 +5,7 @@ require 'feature_spec_helper'
 describe 'Featured works on the home page', type: :feature do
   let!(:user)         { create(:user) }
   let!(:jill_user)    { create(:jill) }
-  let!(:file1)        { create(:featured_file, depositor: user.login, title: ['file title']) }
+  let!(:file1)        { create(:featured_file, depositor: user.login, title: ['file title'], keyword: ["'55 Chet Atkins"]) }
   let!(:file2)        { create(:featured_file, :with_required_metadata, depositor: user.login) }
   let!(:private_file) { create(:private_file, depositor: jill_user.login, title: ['private_document']) }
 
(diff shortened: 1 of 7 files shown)


mtribone:
@awead I say that if it visually matches what is currently in production, we are good to go. Otherwise I could tweak that layout all week!

awead:
@mtribone it doesn't. Here's what's currently in production:
current
Note the bold-face field names are absent.

mtribone:
They just need some `class="sr-only"` on them.

mtribone:
screen shot 2017-10-30 at 2 48 06 pm

awead:
@mtribone ok, but my point was that we could make the same as the show view, which displays those field names instead of hiding them visually. Otherwise, we'd need to create separate renderers for each fields displayed.

mtribone:
@awead okay. so if you'd like to just use the same partial and have them be the same, go for it. Does creator show the bold term or just the other metadata?

awead:
@mtribone it's the depositor, not creator, so we'd probably have to make a new renderer for that. I'm open to suggestions!

olendorf:
looks fine to me

mtribone:
@awead @olendorf I'm good with this for now.

spec/features/featured_work_spec.rb
```diff
@@ -17,11 +17,13 @@
...
     end
 
-    it 'appears as a featured work', js: true do
+    it 'appears as a featured work with translated facets', js: true do
       expect(page).to have_content 'Featured Works'
       within('#featured_container') do
         expect(page).to have_content(file1.title[0])
+        click_link(file1.keyword.first)
       end
+      expect(page).to have_content(file1.title[0])
```
carolyncole, line 26: @awead curious why the second check of title

carolyncole (changes requested):

spec/features/featured_work_spec.rb
awead: @cam156 line 24 clicks on the facet link and we want to make sure it shows the work. Previously, this was not working because the facet wasn't being correctly translated.

PR #1071, opened 2017-11-03
audit_files script

### LINKED ISSUE #1070: Regularly check for missing otherwise messed up content
Given our recent problem with missing content, and also this might help when an upload fails, but have a cron job or something like that that audits the content to check that it is there.

@awead  already wrote a script that does most if not all of this. Also, since it is based on the SHA hash, it would also detect corrupted files?

awead:
Used to query Solr for checksums, and then search for those checksums as
the files in Modeshape's storage layer. This can tell us if files are
missing from Fedora's binary store.

Fixes #1070

### DIFF
--- a/script/audit_files
+++ b/script/audit_files
@@ -1,0 +1,68 @@
+#!/bin/bash
+# ---------------------------------------------------------------------------------------------------------------------
+# Scholarsphere file audit
+#
+# Queries Solr for all the sha256 digests of file sets, parses the output, and then uses it to look in Modeshape
+# to see if the file is actually there. Progress and results are printed to STDOUT.
+# ---------------------------------------------------------------------------------------------------------------------
+
+FEDORA=$1
+SOLR=$2
+
+usage() {
+  echo "
+  audit_files [fedora_path] [solr_host_with_port_and_core]
+
+  Example:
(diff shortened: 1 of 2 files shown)


script/audit_files
```diff
@@ -0,0 +1,68 @@
...
+  Example:
+    audit_files /fcrepo/binary.store http://solr_server:8983/solr/collection1
+
+  Solr defaults to: http://localhost:8983/solr/collection1
+  "
+  exit 0;
+}
+
+audit() {
+  SOLR_URL="${SOLR}/select?q=has_model_ssim%3AFileSet&rows=1000000&fl=digest_ssim&wt=json"
```
carolyncole, line 25: @awead you might want to filter based on model type here.  That way we will not hit the 1 Million soon.  It might also be better to pass in the max rows.

carolyncole (changes requested):

script/audit_files
awead: @cam156 are you suggesting `fq = has_model_ssim:FileSet` instead of `q = has_model_ssim:FileSet` ? Rows has to be high to make sure we're getting all the file sets back.

carolyncole: @awead never mind.  Query is just fine.

PR #1072, opened 2017-11-13
Single Title and Subtitle for Collections

mtribone:
Adds subtitle term to collection form, redefines title method to grab first title, addresses multiple titles, adds subtitle to the primary terms on the form, and adds the subtitle as a new property in the model.

Adds the subtitle field to the collection show page, adds the collection subtitle partial, delegates subtitle to solr in the collection presenter.

Updates create and edit spec test to test for subtitle.

Fixes style error in test for the collection presenter and adds subtitle to the collection form test.

### DIFF
--- a/app/forms/collection_form.rb
+++ b/app/forms/collection_form.rb
@@ -3,6 +3,7 @@
 class CollectionForm < Sufia::Forms::CollectionForm
   attr_reader :current_ability, :request
 
+  self.terms += [:subtitle]
   self.required_fields = [:title, :description, :keyword]
 
   include WithCreator
@@ -17,8 +18,26 @@ class CollectionForm < Sufia::Forms::CollectionForm
     super(model)
   end
 
+  def self.multiple?(field)
+    if field.to_sym == :title
+      false
+    else
+      super
+    end
+  end
+
+  def self.model_attributes(_)
+    attrs = super
+    attrs[:title] = Array(attrs[:title]) if attrs[:title]
+    attrs
+  end
+
   def model_class_name
     'collection'
+  end
+
+  def title
+    super.first || ''
   end
 
   # @return [Array<WorkShowPresenter>]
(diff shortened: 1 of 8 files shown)


mtribone:
screen shot 2017-11-13 at 10 22 07 am

app/forms/collection_form.rb
```diff
@@ -17,10 +18,28 @@ def initialize(model, current_ability, request)
     super(model)
   end
 
+  def self.multiple?(field)
+    if [:title].include? field.to_sym
```
carolyncole: `if field.to_sym == :title` would be an easier way of saying this.

db/schema.rb
```diff
@@ -569,8 +569,8 @@
 
   add_foreign_key "curation_concerns_operations", "users"
   add_foreign_key "mailboxer_conversation_opt_outs", "mailboxer_conversations", column: "conversation_id", name: "mb_opt_outs_on_conversations_id"
-  add_foreign_key "mailboxer_notifications", "mailboxer_conversations", column: "conversation_id", name: "notifications_on_conversation_id_development"
```
carolyncole: This file should be excluded

spec/features/collection/edit_spec.rb
```diff
@@ -76,13 +78,15 @@
...
       expect(page).to have_field 'collection_title', with: original_title.first
+      expect(page).to have_field 'collection_subtitle', with: original_subtitle
       expect(page).to have_field 'collection_description', with: original_description.first
       within('div.collection_date_created') do
         expect(page).to have_content('Published Date')
       end
       expect(page).to have_checked_field('Public')
       expect(page).to have_no_checked_field('Private')
       fill_in 'Title', with: updated_title
+      fill_in 'Subtitle', with: updated_subtitle
```
carolyncole, line 89: There are two check below after the Update button is clicked that I would expect to also see for the subtitle.  Make sure the original subtitle is not present.   Make sure the new subtitle is present.

carolyncole (changes requested):

db/schema.rb
mtribone: oopsies. I forgot to check it out again after I migrated the database.

carolyncole (changes requested):
@mtribone everything looks good except for the schema file.

Issue #1075, opened 2017-11-14
We should capture creator errors


carolyncole:
When I only fill in the surname the error should be captured and sent to the UI.
I should not get an exception that is not captured

![screen shot 2017-11-14 at 7 19 00 am](https://user-images.githubusercontent.com/1599081/32779549-2a81c8cc-c90c-11e7-90a7-4fef69e65c44.png)

PR #1078, opened 2017-11-14
Capturing Alias creation errors

carolyncole:
Captured excption in the actor stack
Display errors in the flash messages
Removed preserve title order since our title is singular

refs #1075

### DIFF
--- a/app/actors/curation_concerns/actors/generic_work_actor.rb
+++ b/app/actors/curation_concerns/actors/generic_work_actor.rb
@@ -9,17 +9,18 @@ module CurationConcerns
   module Actors
     class GenericWorkActor < CurationConcerns::Actors::BaseActor
       def create(attributes)
-        preserve_title_order(attributes)
-        super
+        capture_alias_errors do
+          super
+        end
+      end
+
+      def update(attributes)
+        capture_alias_errors do
+          super
+        end
       end
 
       protected
-
-        # Remove this method once #948 and #949 are resolved.
-        def preserve_title_order(attributes)
-          curation_concern.save
-          curation_concern.title = attributes[:title]
-        end
 
         # Overrides CurationConcerns::Actors::BaseActor to reassign the depositor
         # if the user is depositing on behalf of someone else.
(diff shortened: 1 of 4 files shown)


app/actors/curation_concerns/actors/generic_work_actor.rb
```diff
@@ -31,6 +32,13 @@ def apply_save_data_to_curation_concern(attributes)
           end
           super
         end
+
+        def capture_alias_errors
+          yield
+        rescue AliasManagementService::Error => error
+          curation_concern.errors.add(:creator, :invalid, message: error.message)
+          return false
```
awead, line 40: what happens here? Does the rest of the actor updates/creates continue, or does the whole thing stop? Is the page redirected back to edit or does it continue on to the user dashboard?

carolyncole: @awead I believe that returning false causes the actor stack to stop.  This has nothing to do with the redirect.  The redirect is handled in controller changes I added.

carolyncole:
@awead here is the logic for determining what to do when the actor returns false: https://github.com/samvera/curation_concerns/blob/master/app/controllers/concerns/curation_concerns/curation_concern_controller.rb#L47
and here
https://github.com/samvera/curation_concerns/blob/master/app/controllers/concerns/curation_concerns/curation_concern_controller.rb#L90

app/actors/curation_concerns/actors/generic_work_actor.rb
awead: @cam156 are the controller changes in another PR? I don't see any changes related to a redirect.

carolyncole: I take that back the redirect is handled in the controller see the links I added into the PR.

I just changed the fact that the object errors were displayed in the flash.

carolyncole:
Before the PR
screen shot 2017-11-14 at 2 42 44 pm

awead:
@cam156 looks good! It also looks like it the files have been successfully attached so the files requirement is shown as complete, and the user can just fix the metadata problem. Is the files requirement just always green during edit, or is the form actually checking to see if the work has files?

carolyncole:
@awead I have actually found that the file does not seem to still be attached.   That was just my browser not running the javascript.   

Here is an accurate after the PR:
screen shot 2017-11-15 at 7 00 03 am

awead:
@cam156 ok, so it looks like if there's a creator error, files won't get attached. If that seems right to you, I'm 👍

carolyncole:
@awead it is certainly more than I was trying to accomplish with this ticket.  I was not trying to change the overall actor stack flow, just making sure our error was handled.

PR #1082, opened 2017-11-14
Enable batch edit with Alias resources

awead:
When changing our forms to use Alias objects instead of Agents, the
batch edit form wasn't update to account for the newly added fields.

This updates the batch edit feature test to check that creators can be
removed and added.

### DIFF
--- a/spec/features/batch_edit_spec.rb
+++ b/spec/features/batch_edit_spec.rb
@@ -20,15 +20,38 @@ describe 'Batch management of works', type: :feature do
 
     it 'edits a field and displays the changes', js: true do
       expect(page).to have_content 'Changes will be applied to the following'
+
+      # Update standard fields
       batch_edit_fields.each do |field|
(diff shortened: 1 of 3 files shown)


spec/features/batch_edit_spec.rb
```diff
@@ -20,15 +20,31 @@
...
       expect(page).to have_content 'Changes will be applied to the following'
+
+      # Update standard fields
       batch_edit_fields.each do |field|
         fill_in_batch_edit_field(field, with: "Updated batch #{field}")
       end
+
+      # Update creators
+      first('.remove-creator').click
+      fill_in 'batch_edit_item[creators][1][given_name]', with: 'Updated'
```
carolyncole: @awead I would expect you to need to fill in Surname, not given name.  Is surname already filled out?

carolyncole (commented):
@awead just one question.

carolyncole (approved):

Issue #1041, opened 2017-09-20
Creator search should return results from LDAP


awead:
When adding creators to a work, we need to search beyond the existing users in the system, and include all PSU users in LDAP. Results returned would include PSU id as well as email. Users can select which sources to search, whether existing users or users from LDAP.
![creator-form](https://user-images.githubusercontent.com/312085/30645320-5dd31d4c-9de3-11e7-876a-8d72d64faf52.png)

awead:
Depends on #961

awead:
@cam156 This is to capture the discussion from this morning. I will should be able to add the additional fields to the UI, but might need some help from @mtribone with the dropdown.

mtribone:
When searching the existing users in the new system, does it match on creator name? Which is the display name?

carolyncole:
It matches with Person surname and given name.

PR #1087, opened 2017-11-16
Adding in ldap results to agent query

carolyncole:
refs #1041

### DIFF
--- a/app/controllers/agents_controller.rb
+++ b/app/controllers/agents_controller.rb
@@ -11,7 +11,9 @@ class AgentsController < ApplicationController
   private
 
     def search_results(query)
-      solr_search_results(query)
+      ldap_results = ldap_search_results(query)
+      solr_results = solr_search_results(query)
+      merge_results(solr_results, ldap_results)
     end
 
     # returns solr results in the following format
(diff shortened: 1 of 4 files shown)


app/controllers/agents_controller.rb
```diff
@@ -30,9 +32,62 @@ def solr_search_results(query)
...
+      main_results.sort { |x, y| hash_key(y) <=> hash_key(x) }
+    end
+
+    def hash_key(result)
+      display_name = result['display_name']
+      display_name = display_name.first if display_name.is_a?(Array)
+      display_name = display_name.downcase if display_name.present?
+      email = result['email']
+      email = email.first if email.is_a?(Array)
+      email = email.downcase if email.present?
```
awead: You're doing the same thing with email and display name. Can you put that into a method?

app/controllers/agents_controller.rb
```diff
@@ -30,9 +32,62 @@ def solr_search_results(query)
...
+    #   "sur_name"=>["COLE"],
+    #   "psu_id"=>["cam156"],
+    #   "email"=>["cam156@psu.edu"]}]
+    def ldap_search_results(query)
+      name = Namae::Name.parse(query.upcase)
+
+      email_items = LdapDisambiguate::LdapUser.query_ldap_by_mail(query + '@psu.edu', ldap_attrs)
+      name_items = LdapDisambiguate::LdapUser.query_ldap_by_name(name.given, "#{name.family}*", ldap_attrs)
+      email_items = [] if email_items.blank?
+      name_items = [] if name_items.blank?
```
awead: email_items and name_items are treated identically. How about a method that wraps the response from LdapDisambiguate::LdapUser.query_ldap_by_mail and returns [] if the response is blank?

spec/controllers/agents_controller_spec.rb
```diff
@@ -10,6 +10,56 @@
...
+
+    let(:jamie_l_franks_ldap_entry) do
+      Net::LDAP::Entry.new('uid=qjf1,dc=psu,dc=edu').tap do |e|
+        e[:uid] = ['qjf1']
+        e[:givenname] = ['JAMIE L']
+        e[:sn] = 'FRANKS'
+        e[:mail] = ['qjf1@psu.edu']
+        e[:displayname] = ['JAMIE L FRANKS']
+      end
+    end
```
awead: Can we use some kind of factory here?

carolyncole: @awead good call.  There was already a factory

awead: @cam156 oh nice, I didn't know there was one there.

Issue #1049, opened 2017-10-05
Investigate ordered creators


awead:
Ordering creators the way we do members may have some unintended consequences. The `HasCreators` module is also using `:list_source` which may be overlapping with the `members`.  To make sure the two are independent, we need to test a GenericWork with multiple ordered file sets and multiple ordered creators to make sure the two sets of orderings are truly independent from one another.

awead:
It looks like the creator ordering with an additional indirect container does NOT work. We should probably fall back to the two-predicate approach, unless we feel like digging through the ordering code in AF. I'm afraid that's going to be a waste of time.

PR #1096, opened 2017-11-16
Enable ordered creators with custom predicate, ref #949, #1049

awead:
Implementing an ordered aggregation for creators was not preserving
their order.

The ordered aggregation is removed in favor of a simpler property and a
second custom predicate that maintains the order of creators as a
delimited string of ids.

### DIFF
--- a/app/forms/batch_edit_form.rb
+++ b/app/forms/batch_edit_form.rb
@@ -9,15 +9,61 @@ class BatchEditForm < Sufia::Forms::BatchEditForm
   end
 
   def initialize_combined_fields
-    super
-    permissions = []
-    admin_set_id = ''
-    batch_document_ids.each do |doc_id|
-      work = model_class.find(doc_id)
-      permissions = (permissions + work.permissions).uniq
-      admin_set_id = work.admin_set_id
-    end
-    model.admin_set_id = admin_set_id
-    model.permissions_attributes = permissions.map(&:to_hash).uniq
+    assign_combined_attributes_to_model
+    @names = combinator.names
+    model.creators = combinator.creators
+    model.admin_set_id = AdminSet::DEFAULT_ID
+    model.permissions_attributes = combinator.permissions
   end
+
+  private
+
+    # Assigns each of the combined attributes to {model}, creating an empty array for null attributes.
+    def assign_combined_attributes_to_model
+      combined_attributes.keys.each do |key|
+        model[key] = combined_attributes[key].empty? ? [''] : combined_attributes[key]
+      end
+    end
+
+    def combinator
+      @combinator ||= Combinator.new(batch_document_ids, model_class)
+    end
+
+    def combined_attributes
+      @combined_attributes ||= combinator.attributes(terms - [:creator])
+    end
+
+    class Combinator
+      attr_reader :works
+
+      # @param [Array<String>] ids of works to edit
+      # @param [ActiveFedora::Base] model_class of the works
+      def initialize(ids, model_class)
+        @works = ids.map { |id| model_class.find(id) }
+      end
+
+      # @param [Array<Symbol>] keys for attributes that we want to combine
+      # @return [Hash] with keys of attributes and their combined values from the works
+      # object[key] returns a ActiveTriples::Relation which must be cast to an array in order to flatten it
+      def attributes(keys)
+        results = {}
+        keys.each do |key|
+          results[key] = works.map { |work| work[key].to_a }.flatten.uniq
+        end
+        results
+      end
+
+      def creators
+        works.map(&:creators).flatten
+      end
+
+      def names
+        works.map(&:to_s)
+      end
+
+      def permissions
+        combined_permissions = works.map(&:permissions).flatten
+        combined_permissions.map(&:to_hash).uniq
+      end
+    end
 end
--- a/app/models/batch_edit_item.rb
+++ b/app/models/batch_edit_item.rb
@@ -27,7 +27,8 @@ class BatchEditItem < ActiveFedora::Base
     range.first
   end
 
+  # Creators are ActiveTriples::Relation which must be cast to arrays before they can be flattened
   def creators
-    batch.map(&:creator).flatten
+    batch.map(&:creators).map(&:to_a).flatten
   end
 end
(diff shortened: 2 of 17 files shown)


app/models/concerns/has_creators.rb
```diff
@@ -4,17 +4,33 @@ module HasCreators
...
+    property :creator_list, predicate: PSUL.orderedCreators, multiple: false do |index|
+      index.as :symbol
+    end
+
     def creators=(values)
       values = values.values if values.is_a? Hash
       agent_records = values.map { |v| AliasManagementService.call(v) }
-      super(agent_records)
+      self.creator_list = agent_records.map(&:id).join('##')
+      super(agent_records.map(&:uri))
```
carolyncole: @awead is the map to uri change needed?

app/models/concerns/has_creators.rb
```diff
@@ -4,17 +4,33 @@ module HasCreators
...
+    end
+
+    def creator_ids
+      creators.map(&:id)
```
carolyncole: Should this use the ordered list?

app/models/concerns/has_creators.rb
```diff
@@ -4,17 +4,33 @@ module HasCreators
...
+    end
+
+    def ordered_creator_ids
```
carolyncole: @awead is there a reason we would want the unordered creators?  I would think we would always want the order.

spec/support/helpers/solr_objects.rb
```diff
@@ -4,21 +4,23 @@ def conn
...
+# @param [Collection] collection
+# @param [String] creator
+# @note allows you to save a collection to Solr using a creator as a string
 def save_collection_to_solr(collection, creator)
   allow(Collection).to receive(:find).with(collection.id).and_return(collection)
   allow(collection).to receive(:creator_ids).and_return([creator])
-  hash = collection.to_solr
-  hash['creator_sim'] = creator
-  conn.add hash
+  conn.add(collection.to_solr.merge!('creator_sim' => creator))
```
carolyncole: @awead I realize this is less lines of code, but is it really more readable?

carolyncole (changes requested):

app/models/concerns/has_creators.rb
awead: Yes, it doesn't work with Alias objects, you have to use their uris.

app/models/concerns/has_creators.rb
awead: `creator_ids` could be a bit faster because it's not pulling the `creator_list` predicate and sorting through each one. That may be extra hits to Fedora. It also mimics the `ordered_members` vs. `members` distinction.

spec/support/helpers/solr_objects.rb
awead: It's less overall lines of code to read, and one less variable to have to keep in your head.

app/models/concerns/has_creators.rb
awead: It's being used in `SolrListMigrator`

spec/support/helpers/solr_objects.rb
carolyncole: @awead I would suggest that reading the one line I have to parse it to know what it does, where the multiple lines makes it obvious that er get the solr, change the creator and then send the hash to solr.

carolyncole: @awead this is not boring ruby.  It is fancy ruby I might hesitate to touch later.  Not sure this change is at all related to this PR.

app/forms/batch_edit_form.rb
```diff
@@ -9,15 +9,43 @@ def model_class_name
...
+
+  private
+
+    def build_combined_fields
+      batch_document_ids.each do |doc_id|
+        work = model_class.find(doc_id)
+        combined_attributes.keys.each do |key|
+          combined_attributes[key] = (combined_attributes[key] + work[key].to_a).uniq
+        end
+        combined_creators.push(*work.creators)
```
carolyncole: Should this be ordered?

app/models/batch_edit_item.rb
```diff
@@ -27,7 +27,8 @@ def visibility
     range.first
   end
 
+  # Creators are ActiveTriples::Relation which must be cast to arrays before they can be flattened
   def creators
-    batch.map(&:creator).flatten
+    batch.map(&:creators).map(&:to_a).flatten
```
carolyncole, line 32: Should this be ordered?

app/models/concerns/has_creators.rb
carolyncole: @awead that should then use ordered list since we want to preserve order I would think when we migrate.

app/models/concerns/has_creators.rb
carolyncole: Ok then we need to change all references to creator_ids and creators ordered creators.

spec/support/helpers/solr_objects.rb
awead: @cam156 I pushed up a change. I had a lot of problems with these helpers because it wasn't clear that they were trying to mimic the old form of creators. I added some comments to clear that up, and hopefully the updated method makes that more clear. I spent some time trying to  make a factory that would create a collection or work with the complete form of "old" creators, but that wasn't working out. So I've opted for an extracted method to help clear things up.

carolyncole (commented):
@awead just to be certain I am reading the PR correctly.  When you get the ids you always get them in order, but when you get the object you can get them either in order or in random order?  If that is correct that does not match with how the system currently works.  Creators are expected to be ordered when getting or setting them.

awead:
@cam156 the order of creators are set in `creators=` via the `creator_list` predicate. If we want `creators` to return the ordered set, we'd need to override `creators` which is meta-programmed method and could have unintended consequences. However, we're overriding `creators=` which is also meta-programmed, so we're only making it slightly worse.

carolyncole:
@awead to be certain we are not going to lose our ordering some how it would make sense to override creators.
I could see someone doing this
```
local_creators = object.creators
...
local_creators << new_creator
object.creators = creators
```
That might loose the original ordering.  I also wonder would
`object.creators << new_creator loose the order?` or `object.creators +=[new_creator]`

Yes I may be being a bit paranoid, but this is a large change to be introducing so late.  Maybe we need a full regression test if you really want to change the behavior.

awead:
@cam156 I don't think we can rely on `<<` and `+=` might work, but I would only be certain of using `=` and nothing else. At this point, the only interface to updating the creators should be `creators=`. I can override `creators` and see what happens.

PR #1097, opened 2017-11-16
fixing a few bugs in the migration process

carolyncole:
This PR just hardens the migration process.  
* 1000 was a mistake since we have over 7000 items.
* Random errors cause the process to fail so I added a capture
* The creators for some records were blank.  This could be becuase we were removing them, but it seemed good to handle that case
* Added a throw catch so if the alias is not found we do not clear the creators
* Check the Alias database for the display name if it has not been found in the hash

### DIFF
--- a/app/models/migration/solr_list_migrator.rb
+++ b/app/models/migration/solr_list_migrator.rb
@@ -19,6 +19,8 @@ module Migration
       private
 
         def already_migrated?(object)
+          return true if object.creator_ids.blank? || object.creator_ids[0].blank?
+
           object.creator_ids[0].match(/^[0-9A-F]{8}-[0-9A-F]{4}/i)
         end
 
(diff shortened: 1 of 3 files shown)


app/models/migration/solr_list_migrator.rb
```diff
@@ -19,6 +19,8 @@ def update(object, creator_alias_hash)
       private
 
         def already_migrated?(object)
+          return true if object.creator_ids.blank? || object.creator_ids[0].blank?
+
```
awead, line 23: What about just `object.creators.first.kind_of?(Alias)` for the whole method? That should account for `[]`, `[""]`, as well as strings.

carolyncole: @awead the ids are both strings, and if you look at the creators it tries to load them and starts throwing errors when something is not migrated.

PR #1105, opened 2017-11-20
Changing to use the creator field since that is the field currently p…

carolyncole:
…opulated with the legacy information


awead:
@cam156 LGTM. You can merge it when it's green.

Issue #1109, opened 2017-11-21
LocalUserLookup is not handeling names with commas


carolyncole:
Given a Local user with a display_name of 'HAYOU LI'
We should find that local user if we send
`LocalUserLookup.find_users('HAYOU LI')` or 
`LocalUserLookup.find_users('LI HAYOU')` or 
`LocalUserLookup.find_users('LI, HAYOU')`

Currently only the first two return a user and a new agent is being created that will need to be removed into the future.

PR #1110, opened 2017-11-21
Adding some cleanup to the name parts

carolyncole:
refs #1109


awead (approved):
@cam156 if it matters, `gsub(/[,.]/,''` would remove both commas and periods, but I'm not holding this up for that.

Issue #1104, opened 2017-11-20
Creator facet needs to facet on "Canonical Name" not the alias


olendorf:
When I upload works with two different aliases, there is a facet for each alias. Its probably better to facet on the creators registered name? Still not sure the best word for it.

PR #1112, opened 2017-11-21
Use canonical name for creator facets, ref #1104

awead:
The display was used for facets, which could change from person to
person. To make the facets unique for each individual, the facets are
created from the concatenation of their given and surnames.

### DIFF
--- a/app/indexers/indexes_creator.rb
+++ b/app/indexers/indexes_creator.rb
@@ -13,7 +13,13 @@ module IndexesCreator
 
     def index_creator(solr_doc)
       creator_names = object.creators.map(&:display_name)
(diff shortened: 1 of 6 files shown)


app/indexers/indexes_creator.rb
```diff
@@ -12,8 +12,13 @@ def generate_solr_document
   private
 
     def index_creator(solr_doc)
-      creator_names = object.creators.map(&:display_name)
+      creator_names = object.creators.map { |c| build_name(c) }
```
carolyncole: @awead I think you only want to change the facet not the name int he record also.

PR #1124, opened 2017-11-30
First name initial matches should not match last character of first name

carolyncole:
We should not match Jessica Moyer with A Moyer just becuase the given name ends with the first initial.  Solr is allowing this to come back so we are filtering it when we validate.

### DIFF
--- a/app/models/migration/creator_list.rb
+++ b/app/models/migration/creator_list.rb
@@ -78,15 +78,15 @@ module Migration
         return true if creator == user.login
 
         logger.warn("Matching #{creator} with #{user.display_name} #{user.login}")
-        user_name = parse_name(user.display_name)
+        valid_given_and_sur_name?(user.display_name, creator) || (comparable_name(creator) == comparable_name(user.display_name))
+      end
+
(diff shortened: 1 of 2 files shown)


app/models/migration/creator_list.rb
```diff
@@ -78,15 +78,15 @@ def validate_user(creator, user)
...
-        user_name.family.downcase!
-        creator_name.family.downcase!
-
-        # family name match is valid
-        return true if user_name.family == creator_name.family
+        validates_on_given_and_sur_name(user.display_name, creator) || (comparable_name(creator) == comparable_name(user.display_name))
+      end
 
-        comparable_name(creator) == comparable_name(user.display_name)
+      def validates_on_given_and_sur_name(user_display_name, creator)
```
awead: @cam156 is this returning a true false? rename with ?

carolyncole: True if the given and sur name validate.  Can you give me an example of what you are looking for?  Something like `valid_give_and_sur_name?`

awead: yep, that's it exactly!

awead (approved):

Issue #1016, opened 2017-07-28
Collection Metadata Order


carolyncole:
Order or the metadata on the collection show form should reflect the order of the form.

The size and items should be at the bottom

carolyncole:
![screen shot 2018-04-04 at 10 27 59 am](https://user-images.githubusercontent.com/1599081/38313832-ed012a00-37f2-11e8-9238-43be8a81b364.png)

carolyncole:
Leave Multiple columns but put in the same order as the work

PR #1133, opened 2018-01-04
Moves Collection Creator Field(s) Above Additional Metadata

### LINKED ISSUE #1127: Collection Creator Should be a part of the main fields
Currently creator is under additional fields.
This should move out to the main form like it is for a work.

Creator is not required for collection.

mtribone:
Moves the creator field up in the new/edit form and rearranges the metadata on the collection show page to match work show page. Updates spec test to reflect position change of creator field.

Fixes #1127 and #1016 

screen shot 2018-01-04 at 12 36 29 pm

### DIFF
--- a/app/presenters/collection_presenter.rb
+++ b/app/presenters/collection_presenter.rb
@@ -3,9 +3,10 @@
 class CollectionPresenter < Sufia::CollectionPresenter
   delegate :subtitle, to: :solr_document
 
-  # TODO: Move to Sufia?
   def self.terms
-    super + [:date_modified, :date_uploaded]
+    [:creator, :keyword, :size, :total_items, :resource_type, :contributor,
+     :rights, :publisher, :date_created, :subject, :language, :identifier,
+     :based_near, :related_url, :date_modified, :date_uploaded]
   end
 
   def permission_badge_class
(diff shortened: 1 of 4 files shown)


app/presenters/collection_presenter.rb
```diff
@@ -5,7 +5,9 @@ class CollectionPresenter < Sufia::CollectionPresenter
 
   # TODO: Move to Sufia?
```
awead: Can you remove this TODO? We'll never need to do it.

app/presenters/collection_presenter.rb
```diff
@@ -5,7 +5,9 @@ class CollectionPresenter < Sufia::CollectionPresenter
...
   def self.terms
-    super + [:date_modified, :date_uploaded]
+    [:creator, :keyword, :size, :total_items, :resource_type, :contributor,
+     :rights, :publisher, :date_created, :subject, :language, :identifier,
+     :based_near, :related_url, :date_modified, :date_uploaded]
```
awead, line 9: should we have a test for this? I could go either way...

mtribone: Yeah maybe. I had to copy the whole thing over to adjust the ordering.

PR #1134, opened 2018-01-04
Expire leases and embargoes, fixes #1051

### LINKED ISSUE #1051: Embargo And Lease Cron Job to change permissions
Need to create a script to run through any Embargoes and leases that are out of date and update the works and filesets to the appropriate permissions

awead:
Leases and embargoes and not automatically expired when their dates are
reached.

ExpirationService expires both leases and embargoes on a given work
that are past a given date, by default the date is today. The visibility
of the work and any contained file sets are changed accordingly.

An included rake task with run the service, as well as set a cron job
that runs this task every morning.


awead:
Code Climate is complaining because the `expire_embargoes` and `expire_leases` method look very similar. Well, yes, they do. I could refactor this, but it might start looking like "fancy ruby" and I kind of liked the boring, verbose approach here.

mtribone:
@awead Would code climate complain if you swapped them around? For example, `embargo_expire`?

mtribone:
@awead Ah. Yes. I see. They are quite similar. Can't we do some `if/else` and `||` magic quickly to combine the two methods?

mtribone:
Or just have one visibility expiration method that can be used on any of them?

carolyncole (approved):

PR #1135, opened 2018-01-05
Adds required label to Display Name for Creator

mtribone:
Adds the required label by switching the label to a do end block. Handled slightly differently for creator than other default fields in the form because creator is a special thing.

screen shot 2018-01-05 at 10 38 11 am


awead:
@mtribone you could add a few lines to one of the feature tests that checks to see if the required label is there. I was thinking somewhere in edit_work_spec.rb in the test at line 65.

Issue #1045, opened 2017-09-21
Batch Edit: Setting embargo and lease does not work


carolyncole:
When I got to my works listing
And I choose Multiple works 
and I click Edit Selected
And I click on the collaborators tab
And I choose embargo 
And I choose a date in the future
And I click Save Changes
Then the form should not hang forever (It should set the works to embargo)

Same scenario for lease

carolyncole:
Verified that this was an issue in production by running an edit on stage.

carolyncole:
Here is the stack trace from the server
```
[...]

PR #1137, opened 2018-01-08
Batch edit: setting embargo and lease, ref #1045

awead:
Setting an embargo or lease via batch edit would fail.

This enables the setting/overriding of visibilities, embargoes, and
leases in batch edit mode by relying on the actor stack to perform the
same actions on each work in the batch, as would normally be done if the
work was updated alone.

When the works have existing embargoes or leases, and a different one is
supplied, the newer simply overwrites the existing one. If the
visibility is change, the embargo or lease is expired. This may result
in a slight amount of information loss because the history message will
use the embargoed or leased visibility, and not the actual visibility
value specified in the batch update.


mtribone (commented):
Does the actor update the date modified? Otherwise this looks good to me.

awead:
It's doing the exact same thing that happens when you update a single work, so I would assume yes.

PR #1140, opened 2018-01-09
Remove deleted permissions from works in batch edit

awead:
Batch edit was unable to remove permissions when they were removed via
the UI.

Using a separate class to manage the update process for each work in the
batch, permissions are recast using the correct ids for each permission
currently assigned to the work.


carolyncole:
@awead we estimated the ticket as a 5 so there is plenty of time to create a best case solution.

app/controllers/batch_edits_controller.rb
```diff
@@ -48,24 +34,65 @@ def form_class
...
-    def visibility_status(curation_concern)
-      selected_visibility = work_params.fetch(:visibility, nil)
-      return false unless selected_visibility
-      curation_concern.visibility != selected_visibility
-    end
+    # Updates terms, permissions, and visibility for a given object in a batch.
+    # Note: Permissions and visibility are *always* copied down to any contained FileSet objects.
+    #       There is no UI option presented to the user to prevent this, unlike the option that
+    #       is present when changing permissions on a single work.
+    class BatchTask
```
carolyncole: @awead What is Batch Task trying to accomplish?  Did you split this out to make it easier for testing?  Is it really a ConcernBatchUpdate?

app/controllers/batch_edits_controller.rb
```diff
@@ -48,24 +34,65 @@ def form_class
...
+        params
+      end
+
+      def visibility_changed?
+        selected_visibility = params.fetch(:visibility, nil)
+        return false unless selected_visibility
+        initial_visibility != selected_visibility
+      end
+
+      def save_changes
```
carolyncole: @awead I should have commented on your previous PR.  This is not save_changes, but save_embargo_and_lease

app/controllers/batch_edits_controller.rb
```diff
@@ -48,24 +34,65 @@ def form_class
...
+        VisibilityCopyJob.perform_later(curation_concern) if visibility_changed?
+        InheritPermissionsJob.perform_later(curation_concern) if permissions_without_deletes.present?
+      end
+
+      # Removes any deleted permissions
+      def actor_params
+        return params unless params.key?(:permissions_attributes)
+        new_permissions = permissions_without_deletes
+        new_permissions.push('type' => 'person', 'name' => curation_concern.depositor, 'access' => 'edit')
+        params[:permissions_attributes] = new_permissions
```
carolyncole: Could you possibly just figure out the adds and the deletes here instead of overwriting?

app/controllers/batch_edits_controller.rb
awead: @cam156 it's updating an individual work in the batch. I moved it out to its own class to better organize it.

app/controllers/batch_edits_controller.rb
awead: Unfortunately not. If we wanted to remove the individual permissions, we'd need to pass each one in from the form. The form is only rendering one of them, so we'd need to rewrite the batch edit UI and JS so that it would display the one common permission, but also render hidden fields for each of the other permission ids.

carolyncole: @awead I am suggesting looking at the current permissions on the work and looking at the data from the form.  Not just looking at the data returned from the form.

app/controllers/batch_edits_controller.rb
carolyncole: So would it be ConcernUpdateTask or WorkUpdateTask?

app/services/batch_item_update_service.rb
```diff
@@ -0,0 +1,86 @@
...
+  attr_reader :curation_concern, :params, :current_user, :initial_visibility
+
+  # @param [String] id
+  # @param [ActionController::Parameters] params from the BatchEditsController
+  # @param [User] current_user from the BatchEditsController
+  def initialize(id, params, current_user)
+    @curation_concern = ActiveFedora::Base.find(id)
+    @params = params
+    @current_user = current_user
+    @initial_visibility = curation_concern.visibility
```
carolyncole: @awead this confused me.  I thought you were saving this off to make sure if there was changes, but you are really only comparing it in one place visibility_changed.  Is there really a need to copy this off?

app/services/batch_item_update_service.rb
```diff
@@ -0,0 +1,86 @@
...
+    def permissions_with_ids
+      separated_permissions.first
+    end
+
+    def permissions_without_ids
+      separated_permissions.last
+    end
+
+    def separated_permissions
+      @separated_permissions ||= params.fetch(:permissions_attributes, {}).values.partition { |h| h.key?('id') }
```
carolyncole, line 67: @awead how do you know that they is more than one item in this list?  It seems like permisions_with_ids and permissions_without_ids could end up being the first (only) item in the list.

app/services/batch_item_update_service.rb
```diff
@@ -0,0 +1,86 @@
...
+    InheritPermissionsJob.perform_later(curation_concern) if permissions_changed?
+  end
+
+  private
+
+    # Adjusts the permissions parameters so that any permissions with ids are changed to match the
+    # permissions on the curation_concern.
+    def actor_params
+      return params unless permissions_changed?
+      return params unless has_incorrect_ids?
```
carolyncole: This double negative was a bit hard to follow

carolyncole (commented):
I like how you broke out BatchItemUpdateService.  The service itself has a few minor read issues.

app/services/batch_item_update_service.rb
awead: Huh, yeah, I don't know why that was there. It must have been some vestigial code as this class has been evolving over the last week.

app/services/batch_item_update_service.rb
awead: I had tried using an OR operator, but that didn't seem to work. Maybe `if !permissions_changed? && !has_incorrect_ids?` ?

app/services/batch_item_update_service.rb
awead: It's a function of partition: https://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-partition. There are always two arrays returned.

carolyncole (approved):

PR #1147, opened 2018-01-18
adding in DOI service using ezid

carolyncole:
Adding a Check box into the Work New & Edit Screen to create a DOI on save.  This form does not show on the Collection or the Batch upload, which could be a next step. Refs #1147 
screen shot 2018-01-18 at 9 55 43 am


carolyncole:
If DOI has already been created the checkbox is disabled:
screen shot 2018-01-18 at 10 14 34 am

The DOI is stored in the identifier field:
screen shot 2018-01-18 at 10 15 38 am

carolyncole:
@awead I am having trouble figuring out what coveralls is barking about.  Any thoughts?

awead:
@cam156 it looks fine now. Did you re-run it? Occasionally, I see this sort of thing too. It usually goes away when I push an updated branch. It could have something to do with the split builds.

app/services/curation_concerns/doi_service.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module CurationConcerns
+  class DOIService
```
awead: Why put it under the CurationConcerns namespace? I didn't see an existing DOIService class there.

awead (commented):
Looks great... just had one small question.

app/services/curation_concerns/doi_service.rb
carolyncole: @awead I was thinking it was a service for concerns, but I do not particularly love this decision.

awead: @cam156 A service seems fine to me, I was just thinking it didn't need to be under the `CurationConcerns` module namespace.

PR #1148, opened 2018-01-19
Copies Value of ORCID ID from Profile to Agent ORCR ID

### LINKED ISSUE #1101: ORCID ID from profile should be part of the creator
The user profile contains an ORCID, which should be part of the agent if it is filled in.
Need a way to copy existing user ORCIDs to agents.  
Need to hook in profile edit of ORCID to agent update.

mtribone:
Creates new method in the user controller and checks to see if an orcid from User is present and the User login equals the Agent PSU ID and assigns the value from orcid to orcid_id.

Sets a variable for agent, makes sure that an orcid and agent is present and assigns the orcid from user to agent. Saves agent. Adds permitted parameters.

Tests that there is a redirect after profile update, reloads the user ORCID and makes sure that it is present and equals the correct value. Assigns it to the Agent ORCID ID and saves, then checks to make sure that the value is correct.

refs #1101

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -4,6 +4,7 @@ class UsersController < ApplicationController
   include Sufia::UsersControllerBehavior
 
   before_action :linkedin_url, only: :show
+  after_action :copy_orcid_id, only: :update
 
   def linkedin_url
     @linkedin_url ||= format_linkedin_url
@@ -23,6 +24,14 @@ class UsersController < ApplicationController
     end
     respond_to do |format|
       format.json { render json: users.first(20).to_json }
+    end
+  end
+
+  def copy_orcid_id
+    agent = Agent.where(psu_id: @user.login).first
+    if @user.orcid.present? && agent.present?
+      agent.orcid_id = @user.orcid
+      agent.save
     end
   end
 
--- a/spec/controllers/users_controller_spec.rb
+++ b/spec/controllers/users_controller_spec.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe UsersController, type: :controller do
+  let(:user) { create(:user, login: 'abc123') }
+  let!(:agent) { create(:agent, psu_id: user.login) }
+
+  before { sign_in user }
+  routes { Sufia::Engine.routes }
+
+  describe '#update' do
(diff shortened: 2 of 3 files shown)


mtribone:
@awead Please scrutinize this PR as I am doing things with ORCID IDs and Agent.

app/controllers/users_controller.rb
```diff
@@ -26,6 +26,19 @@ def index
     end
   end
 
+  def update
+    super
+    copy_orcid_id
+  end
```
awead: would a `after_update` hook be better here instead of overriding `update`?

spec/controllers/users_controller_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+  describe '#update' do
+    it 'updates the orcid and copies the orcid to the agent' do
+      put :update, user: { 'orcid' => 'http://orcid.org/1234-5678-9012-0000', 'twitter_handle' => '',
+                           'facebook_handle' => '', 'googleplus_handle' => '' },
+                   id: user.login
+      expect(response).to be_redirect
+      user.reload.orcid
+      expect(user.orcid).to eq 'http://orcid.org/1234-5678-9012-0000'
+      agent.orcid_id = user.orcid
+      agent.save
```
awead: The above two lines need to be removed. The assertion you're testing on line 22 is being explicitly set here, so the test really isn't doing anything. I think you want to just assert that the agent's orcid id is set correctly, which the controller should be doing for you. Changing line 7 to `let!(:agent)` will create the agent before the test is run, and then the controller should be able to find it when it runs. You could also add another assertion above line 14, before the controller call, that verifies `agent.orcid_id` is nil.

awead (changes requested):
The spec test needs a little work, but otherwise this is a fine approach.

carolyncole:
@mtribone this is one part of #1101.  It hooks into the edit and copies. We would still need "Need a way to copy existing user ORCIDs to agents".  Could you change the comment to refs?

app/controllers/users_controller.rb
mtribone: So in the User controller, instead of overriding update, I can just simply state `after_update copy_orcid_id` ?

awead: @mtribone yep!

awead (approved):

PR #1149, opened 2018-01-22
forcing UTF8 on all text

carolyncole:
Solr has been returning an illegal UTF to Ascii error, which we have been missing becuase it happens in the background.  This will eliminate the error


awead:
@cam156 how pernicious is the error? We could opt for `try_convert` and still keep the unconverted string if something goes awry, or catch the exception. That way, we wouldn't halt the whole process.

carolyncole:
@awead I believe solr is sending us utf8, but we are pretending it is ascii.  I will try try_convert and see if that fixes it up too.

carolyncole:
@awead there is no method try_convert.  I will add a rescue.

carolyncole:
@awead ready for another review

Issue #1038, opened 2017-09-19
BatchEditsController passes user instead of ability


awead:
The BatchEditsController uses passes the current user as the parameter in `#edit`, when it should be the current ability. Although the controller is passing the wrong class, and BatchEditForm expects the second parameter to be the ability and not user, the form still appears to work. However, we should correct this as some point.

carolyncole:
Change this line to pass current_ability: https://github.com/psu-stewardship/scholarsphere/blob/develop/app/controllers/batch_edits_controller.rb#L13

PR #1151, opened 2018-01-22
Pass current_ability to form instead of current_user, ref #1038

awead:
BatchEditsController was passing the current_user to the form instead of
the current_ability, which the form expects.

This wasn't causing any errors since the two classes responded similarly
enough, but we want to be consisted and conform to the form's interface
expectations.


carolyncole (commented):
Is there any test that could be written?

awead:
@cam156 I don't know. The test for the form is already passing ability in. So if anything, this just aligning the controller more with the test.

carolyncole (approved):

Issue #1163, opened 2018-01-25
LDAP tests failing locally


awead:
The name disambiguation tests are failing because LDAP is returning capitalized results and the test is expecting all caps.

PR #1165, opened 2018-01-25
Updating local LDAP tests, ref #1163

awead:
The responses from our LDAP server have changed slightly, making the
NameDisambiguation test fail.

We've updated the tests to correspond with the new output from LDAP. One
test is marked pending because the user no longer exists.

### DIFF
--- a/spec/services/name_disabiguation_service_spec.rb
+++ b/spec/services/name_disabiguation_service_spec.rb
@@ -9,7 +9,7 @@ describe NameDisambiguationService, unless: travis? do
     let(:name) { 'Thompson, Britta M' }
 
     it 'finds the user' do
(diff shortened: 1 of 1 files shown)


spec/services/name_disabiguation_service_spec.rb
```diff
@@ -66,6 +66,7 @@
     let(:name) { 'Suzanne Paterno *' }
 
     it 'cleans the name' do
+      pending 'User is no longer in LDAP'
```
carolyncole: The test was really testing removing the * that Sue Paterno had next to her name.  We could just use one of us with a star

carolyncole (changes requested):
Looks good except for the pending.  Added a suggestion for removing the pending.

awead:
@cam156 amended.

carolyncole (approved):

Issue #1167, opened 2018-01-25
Scholarsphere and DOIs content types don't map well


olendorf:
Until this is fixed I can fix them manually. Still labeled it as high.

ScholarSphere      ->    DOI
---
Article  -> Text
Audio -> Sound
Book ->
Capstone Project -> Text
Conference Proceeding -> Text
Dataset -> Dataset
Image -> Image
Journal -> Text
Map or Cartographic Material -> Image
Part of  Book -> Text
Poster -> Audiovisual
Presentation -> Audiovisual
Project -> Other
Report -> Text
Research Paper -> Text
Software or Program Code -> Software
Video -> Audiovisual
Other -> Other

awead:
@olendorf Scholarsphere's resource types can be multiple. I don't see if they can be in EzId, such as separating them with commas. If they have to singular, we can just choose the first.

olendorf:
@awead no just singular, so yeah go with the first

awead:
@olendorf I've got a fix in for it. We could do a hotfix release early next week, if you want.

olendorf:
Sounds good

PR #1169, opened 2018-01-25
Map our resource types to those in EzId, ref #1167

awead:
Our resource types are not compatible with those used in EzId.

When sending DOI requests to EzId, we change the resource type sent to
EzId to match their vocabulary.

### DIFF
--- a/app/services/doi_service.rb
+++ b/app/services/doi_service.rb
@@ -3,6 +3,27 @@
 class DOIService
   attr_accessor :client, :handle
 
+  EZID_RESOURCE_TYPES = {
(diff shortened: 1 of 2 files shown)


carolyncole (commented):
@awead I wonder how this works for a collection.  Right now I believe the resource type is empty for a collection.

awead:
@cam156 you'd get a null or empty resource type. I don't know if EzId would just ignore the null key, so I updated the PR to only add the resource type if there is one.

carolyncole (approved):

PR #1186, opened 2018-03-14
Upgrading to solr 7

carolyncole:
Modifying the configs for solr 7 and upgrading solr_wrapper to run solr 7

### DIFF
--- a/.solr_wrapper
+++ b/.solr_wrapper
@@ -1,5 +1,5 @@
 # .solr_wrapper
-version: 5.3.1
+version: 7.1.0
 port: 8983
 instance_dir: tmp/solr-development
 download_dir: tmp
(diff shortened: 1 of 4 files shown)


awead:
@cam156 here's the commit I put it. I was going to put this in a PR with the full text search, but you can look at it here: https://github.com/psu-stewardship/scholarsphere/commit/8b3b9ee5037c8cb2af7f520530b805f044ba1a54

It makes the same changes to schema and solrconfig, but also updates the rest of the config files and removes the ones that aren't being used.

carolyncole:
@awead Did you want to pull it out so that everyone can use it?  I would be happy to close my PR if you do that.

awead:
@cam156 sure... hold on one sec.

PR #1190, opened 2018-03-15
Adds the Contents of a README file to the Work Show

mtribone:
Adds a readme method that selects files from the file_set_presenter that start with some form of readme. Tests the readme method for txt, md, and blank files, and updates the work show view so that if there is a presenter.readme display the contents of the readme using the redcarpet gem otherwise print a statement asking for a readme.


mtribone:
Work show page without a README


screen shot 2018-03-20 at 9 39 50 am

mtribone:
Work show page with a README


screen shot 2018-03-20 at 9 44 42 am

app/views/curation_concerns/base/_work_readme.html.erb
```diff
@@ -0,0 +1,8 @@
+<h2>README</h2>
+<section class="readme">
+  <% if presenter.readme %>
+    <%= raw(presenter.readme) %>
+  <% else %>
+    <p>How about adding a README file? We currently accept plain text (README.txt) and markdown (README.md) files.</p>
```
awead: i18n would be better here.

app/views/curation_concerns/base/_work_readme.html.erb
```diff
@@ -0,0 +1,8 @@
+<h2>README</h2>
```
awead: i18n as well, probably...

awead (changes requested):

awead:
@mtribone I'm kinda on the fence about the feature test. It would be definitely be nice. If we didn't do the feature, we'll need to add a step to our scripted tests that includes uploading a README file. @cam156 thoughts? We can still poke around trying to get the feature test to work.

mtribone:
@cam156 @awead I'm down with getting the feature test to work, but I might work on something else for a bit to cleanse the mind.

awead (approved):

PR #1191, opened 2018-03-16
Removing unused traits and factories

awead:
Work done during a Friday refactoring session.

### DIFF
--- a/spec/factories/file_sets.rb
+++ b/spec/factories/file_sets.rb
@@ -35,13 +35,6 @@ FactoryGirl.define do
       end
     end
 
-    trait :pdf do
-      title ['fake_document.pdf']
-      before(:create) do |fs|
-        fs.title = ['Fake PDF Title']
-      end
-    end
-
     trait :with_file_size do
       after(:build) do |fs|
         allow(fs).to receive(:file_size).and_return('1234')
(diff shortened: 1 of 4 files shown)


spec/models/user_mailer_spec.rb
```diff
@@ -23,7 +23,7 @@
     let(:message)      { described_class.stats_email(1.day.ago, DateTime.now) }
     let(:mock_service) { double }
     let(:csv)          { "a,b,c\nd,e,f\n" }
-    let!(:generic_work) { create :work, :with_pdf }
+    let!(:generic_work) { create :work }
```
carolyncole, line 26: @awead you are removing with pdf here.  Is there a reason?

spec/factories/file_sets.rb
```diff
@@ -35,13 +35,6 @@
       end
     end
 
-    trait :pdf do
```
carolyncole, line 38: This was used until you removed it.

carolyncole (changes requested):

spec/models/user_mailer_spec.rb
awead: @cam156 Whether or not the work has a pdf is irrelevant to the test. Omitting it has no effect on the outcome.

spec/factories/file_sets.rb
awead: @cam156 If the `:with_pdf` work trait isn't needed, then this trait isn't needed either.

carolyncole (approved):

PR #1197, opened 2018-03-20
Use zip format in url for download

awead:
This prevents the zip download service from running if a thumbnail is
being rendered, or if the user is downloading the file from a specific
file set.

### DIFF
--- a/app/controllers/downloads_controller.rb
+++ b/app/controllers/downloads_controller.rb
@@ -15,9 +15,7 @@ class DownloadsController < ApplicationController
     end
 
     def load_file
-      super unless asset.is_a? GenericWork
-
-      zip_service = WorkZipService.new(asset, current_ability, work_directory)
+      return super if params['file'] == 'thumbnail' || asset.is_a?(FileSet)
       zip_service.call
     end
 
@@ -25,5 +23,11 @@ class DownloadsController < ApplicationController
       directory = File.dirname CurationConcerns::DerivativePath.derivative_path_for_reference(asset.id, 'zip')
       FileUtils.mkpath directory
       directory
+    end
+
+  private
+
+    def zip_service
+      WorkZipService.new(asset, current_ability, work_directory)
     end
 end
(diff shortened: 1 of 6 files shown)


app/controllers/downloads_controller.rb
```diff
@@ -15,15 +15,22 @@ def authorize_download!
     end
 
     def load_file
-      super unless asset.is_a? GenericWork
-
-      zip_service = WorkZipService.new(asset, current_ability, work_directory)
-      zip_service.call
+      if params['format'] == 'zip'
```
carolyncole: @awead how about a guard clause here?

app/services/work_zip_service.rb
```diff
@@ -32,7 +32,7 @@ def initialize(work, ability, zip_directory = 'tmp/')
   def call
     zipfile_name = File.join(zip_directory, "#{work.title.first.parameterize('_')}.zip")
 
-    Zip::File.open(zipfile_name, Zip::File::CREATE) do |_zipfile|
+    Zip::File.open(zipfile_name, Zip::File::CREATE) do |zip_file|
```
carolyncole, line 35: Good catch.  Darn refactoring...

carolyncole (changes requested):

carolyncole:
@awead I'm concerned about the testing for the WorkZipService.  Obviously if the  test was good it should have caught the issue.  Did you want me to try and figure out what is messed up in the tests?

app/controllers/downloads_controller.rb
```diff
@@ -15,15 +15,22 @@ def authorize_download!
     end
 
     def load_file
-      super unless asset.is_a? GenericWork
-
-      zip_service = WorkZipService.new(asset, current_ability, work_directory)
-      zip_service.call
+      if params['file'] == 'thumbnail' || asset.is_a?(FileSet)
```
carolyncole: A guard clause would be better here

carolyncole (approved):

Issue #1142, opened 2018-01-10
Accessibility of Visibility and Sharing With Section When Editing a Fileset


mtribone:
When I select to edit a fileset that is part of a work and select the permissions tab, the "Visibility" section doesn't appear to be aligned with the "Visibility" section of the Work form and has missing and multiple form labels. Also the table with person or groups that you are sharing with has orphaned form labels.

mtribone:
![screen shot 2018-01-10 at 10 10 10 am](https://user-images.githubusercontent.com/4163828/34779546-89165e70-f5ee-11e7-98d7-27cecd3146a9.png)

mtribone:
![screen shot 2018-01-10 at 10 10 27 am](https://user-images.githubusercontent.com/4163828/34779553-904436cc-f5ee-11e7-87bf-e3a9d40a8b5c.png)

mtribone:
And a couple contrast errors

![screen shot 2018-01-10 at 10 11 56 am](https://user-images.githubusercontent.com/4163828/34779624-c7e14f98-f5ee-11e7-884e-31da078cdf40.png)

carolyncole:
We should update the file set form to use the work for instead of it' own form.  Then retest for accessibility.

PR #1198, opened 2018-03-21
Accessibility of Fileset Permissions Visibility 

mtribone:
Moves the labels and radio buttons into an unordered list and makes adjustments to the styles.

Overrides the form permissions embargo and lease partials from Curation Concerns to address the multiple labels and accessibility.

Removes <small> in header with <p> and a help class to address HTML and accessibility for color contrast. Adds i18n for help block and representative media statement.

refs #1142

### DIFF
--- a/app/views/curation_concerns/base/_form_permission_embargo.html.erb
+++ b/app/views/curation_concerns/base/_form_permission_embargo.html.erb
@@ -1,0 +1,8 @@
+<%# Overrides the embargo permission partial because of too many nested labels. We've moved the label to
+  # _form_permission.html.erb and have removed the nesting. One label for the radio button and the following
+  # elements can have their labels for the inputs. %>
+<div class="form-inline">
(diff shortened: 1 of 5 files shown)


app/views/curation_concerns/base/_form_permission_embargo.html.erb
```diff
@@ -0,0 +1,5 @@
+<div class="form-inline">
```
awead, line 4: Can you put in a `<%# comment %>` above saying which partial we're overriding and why?

app/views/curation_concerns/base/_form_permission_lease.html.erb
```diff
@@ -0,0 +1,5 @@
+<div class="form-inline">
```
awead, line 4: Same here.

awead (approved):

Issue #624, opened 2017-02-08
As a user I would like to download an entire collection 


olendorf:
Probably just do it as a zip file of the contents.

carolyncole:
One zip file for all works in the collection.  Zipping up the zip file for each work in the collection.  Dependant on #625

carolyncole:
After the button is clicked the user will get feedback to say that the zip is being created, and then data will then be returned when it is done.  Help information beside the button saying this may take some time.

carolyncole:
We may want this to be a different route than download to have the ability to block robots.

PR #1199, opened 2018-03-22
Downloading an entire collection, ref #624

awead:
Using WorkZipService as a basis, create a zip file from a collection containing individual zip files of each work.

Refactors the work zip service so the two can share more code, and renames the spec test to match the Ruby file name.


app/controllers/downloads_controller.rb
```diff
@@ -28,6 +28,15 @@ def work_directory
...
 
     def zip_service
-      WorkZipService.new(asset, current_ability, work_directory)
+      case asset
+      when GenericWork
+        WorkZipService.new(asset, current_ability, work_directory)
+      when Collection
+        CollectionZipService.new(asset, current_ability, work_directory)
+      else
+        raise ZipServiceError, "#{asset.class} cannot be downloaded as a zip file"
```
carolyncole, line 37: 👍

app/services/work_zip_service.rb
```diff
@@ -13,15 +13,15 @@
 #   zip_file_name = service.call
 #
 class WorkZipService
```
carolyncole, line 15: @awead I wonder if this should be renamed now that it is more generic.  Resource?

app/services/work_zip_service.rb
```diff
@@ -30,23 +30,27 @@ def initialize(work, ability, zip_directory = 'tmp/')
   #
   # @return [String] zip file name
   def call
-    zipfile_name = File.join(zip_directory, "#{work.title.first.parameterize('_')}.zip")
+    zipfile_name = File.join(zip_directory, "#{resource.title.first.parameterize('_')}.zip")
 
     Zip::File.open(zipfile_name, Zip::File::CREATE) do |zip_file|
-      work.file_sets.each do |file_set|
-        add_file_set_to_zip(zip_file, file_set)
+      resource.send(child_files).each do |file_set|
```
carolyncole: @awead I'm not in love with this.  How about just allowing child files to return the list to process?

app/services/work_zip_service.rb
awead: @cam156 sure. We could also have the one class handle both collections and generic works instead of having two differently-named services.

awead: @cam156 On second thought, `WorkZipService` is intended to accept a `GenericWork`, so I think the name is apt. We could define a `ResourceZipService` as an abstract class and then create work and collection services that extend it, but that seems a bit much maybe?

Issue #992, opened 2017-07-14
Allow for querying with aliases to match all the records for that user


carolyncole:
Update the solr records with Alias field for creator in collection and work
Add alias field into the search fields. Dependant on #991

carolyncole:
We need to decide if we are going to allow fuzzy matches or exact matches

carolyncole:
Both Rob and Robert should return values for Robert Olendorf.

carolyncole:
Current State: 
Carolyn Cole linked to Carolyn Munson as a alias
Work1 - Display name Carolyn Cole
Work2 - Display name Carolyn Munson
Search for Cole returns work1 & work2
Search for Munson returns only work2 (should return work1)

PR #1202, opened 2018-03-23
Adding canonical name to alias

carolyncole:
This allows for user to query via the canonical name and find the aliases
A record with a display name "Carolyn Munson" tied to the Carolyn Cole agent will be found when the user queries for "cole"
The opposite is not true though. With a record with a display name "Carolyn Munson" and a record of with a display name of "Carolyn Cole", tied to the Carolyn Cole agent
  only the "Carolyn Munson" record would be  be found when the user queries for "munson"

refs #992


carolyncole:
@awead this looks ok too.

PR #1204, opened 2018-03-23
Install Niftany

awead:
Installs the [niftany gem](https://github.com/psu-libraries/niftany) to use for our Rubocop configurations. Includes
at update to Rubocop, and small set of autocorrects, and an update to
our todo list to account for the changes to the gem.

Cleans up our erb files using erb-lint with a config file managed in
Niftany.

Renames the rubocop stage in Travis to niftany, and runs both rubocop
and erb-lint to check for errors.


app/views/_toolbar.html.erb
```diff
@@ -3,49 +3,49 @@
...
             <% end %>
             <ul class="dropdown-menu">
-              <li><%= link_to t("sufia.toolbar.works.my"), sufia.dashboard_works_path %></li>
+              <li><%= link_to t('sufia.toolbar.works.my'), sufia.dashboard_works_path %></li>
 
               <% classification = CurationConcerns::QuickClassificationQuery.new(current_user) %>
               <% classification.each do |concern| %>
                   <li><%= link_to(
-                              "#{t("sufia.toolbar.works.new")}",
+                            t('sufia.toolbar.works.new').to_s,
```
mtribone, line 29: Does this need to use the method `to_s` ?

.erb-lint.yml
```diff
@@ -0,0 +1,51 @@
+---
+# Uses File.fnmatch? for comparison
+exclude:
+  - '*vendor/*'
+linters:
+  ClosingErbTagIndent:
+    enabled: true
+  FinalNewline:
```
mtribone, line 8: This becomes true on line 12

app/views/collections/_form.html.erb
```diff
@@ -10,9 +10,9 @@
             '#extended-terms',
             class: 'btn btn-default additional-fields',
             data: { toggle: 'collapse' },
-            role: "button",
-            'aria-expanded'=> "false",
-            'aria-controls'=> "extended-terms" %>
+            role: 'button',
+            'aria-expanded' => 'false',
+            'aria-controls' => 'extended-terms' %>
```
mtribone, line 15: It missed the hash rocket here unless it corrected it father down the list.

mtribone (commented):
Just a few things. I also wonder about some of the string interpolation without the curly braces, but I guess we'll spin it up and see if it works.

.erb-lint.yml
awead: @mtribone ah! good catch. We want this enabled, I believe.

app/views/_toolbar.html.erb
awead: Yes, because it's no longer in double-quotes with `#{}`. We could change this back if we like the other way.

app/views/collections/_form.html.erb
awead: I think hash rockets are allowed if the key is a string

PR #1207, opened 2018-03-27
Create spaces in text fields for indexing

awead:
Uses Solr's MappingCharFilterFactory to replace underscores and other
characters with spaces so that search on their parts yields search
results.

Text and string fields are copied to the all_text_timv field, removing
the suggest fields, which aren't being used.

Note: This will require reindexing Solr for the fix to take effect.


spec/controllers/catalog_controller_spec.rb
```diff
@@ -168,6 +178,7 @@
         expect(assigns(:document_list).count).to be(3)
       end
       it 'finds a file by depositor in advanced search' do
+        pending 'Advanced search is currently not used'
```
carolyncole: @awead this used to pass, so why does it need to be pending now?  We do allow users to use OR in thier query.  It seems bad to remove functionality.

carolyncole (changes requested):
Make sure all the tests still pass.  Marking another test as pending seem like a bad idea.  If something was broken we should fix it.

carolyncole (approved):

PR #1213, opened 2018-04-03
Adds the Collection Info to the Work in the My Works List

mtribone:
Reuses the collection presenter from the work show page to display an inline list of collections. Adding the if statement back into the block just like in the work show view so that works without collections do not show the part of empty statement.

Moves the collection information to the metadata details under the disclosure triangle. Adds the description terms to the sufia yml file.

### DIFF
--- a/app/views/my/_index_partials/_list_works.html.erb
+++ b/app/views/my/_index_partials/_list_works.html.erb
@@ -23,10 +23,6 @@
           <a href="#" class="small" title="Click for more details">
             <span id="expand_<%= document.id %>" class="glyphicon glyphicon-chevron-right" aria-hidden="true"></span>
           </a>
-
-          <br />
-          <%= render_collection_links(document) %>
-
         </div>
       </div>
     </div>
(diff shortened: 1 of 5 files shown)


mtribone:
screen shot 2018-04-03 at 2 39 15 pm

mtribone:
screen shot 2018-04-03 at 2 39 58 pm

app/views/my/_index_partials/_list_works.html.erb
```diff
@@ -46,11 +42,11 @@
   </td>
 </tr>
 <tr id="detail_<%= document.id %>">
-  <td colspan="6">
+  <td colspan="5">
     <dl class="expanded-details row">
-      <dt class="col-xs-3 col-lg-2">Creator:</dt>
+      <dt class="col-xs-3 col-lg-2"><%= t('sufia.dashboard.my.creator') %></dt>
```
carolyncole, line 47: 👍

app/views/my/_index_partials/_list_works.html.erb
```diff
@@ -60,6 +56,20 @@
         <% end %>
         <%= t('sufia.dashboard.my.collection_list.users') %> <%= document.edit_people.join(', ') %>
       </dd>
+        <% collection_presenters = presenter.collection_presenters %>
+        <% if collection_presenters.blank? %>
```
carolyncole: use `<% if collection_presenters.present? %>` instead of if and else

app/views/my/_index_partials/_list_works.html.erb
```diff
@@ -60,6 +56,20 @@
...
+        <% else %>
+          <dt class="col-xs-3 col-lg-2"><%= t('sufia.dashboard.my.collection') %></dt>
```
carolyncole: I wonder if this should be a partial.  I'm not sold on this, just seems a bit busy for in the middle here.

mtribone: @cam156 I think that's a good idea.

carolyncole (approved):

PR #1217, opened 2018-04-04
Require zip

awead:
Downloading zips from our server environment was failing because the Zip
module was not being loaded. Because the error was not occurring during
local development, the issue is likely due to the eager loading in
development that is not present in production Rails environments.

### DIFF
--- a/app/services/work_zip_service.rb
+++ b/app/services/work_zip_service.rb
@@ -12,6 +12,8 @@
 #
 #   zip_file_name = service.call
 #
+require 'zip'
+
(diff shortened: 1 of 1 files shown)


carolyncole:
@awead did you deploy this out to qa to see if it fixes the issue?  I tested QA and it seems to be working.

awead:
@cam156 I made the change directly in QA at first to see if that was the problem, then I made PR. Another idea I had was to get the development environment to have the error, which would catch these errors early.

PR #1247, opened 2018-05-21
Update README to remove references to sample yml

little9:
The sample configs aren't present in the repository anymore.


carolyncole:
@little9 Sorry!  They are present, but they got moved to config sample:  https://github.com/psu-stewardship/scholarsphere/tree/develop/config/sample

PR #1252, opened 2018-05-23
Unbold Description Terms and Bold Description Definitions

mtribone:
Overrides Bootstrap styles system-wide for description lists, so that the term is no longer bold and the definition is now bold. This one's for @olendorf because we like him.


mtribone:
screen shot 2018-05-23 at 6 30 24 pm

mtribone:
screen shot 2018-05-23 at 6 30 45 pm

mtribone:
screen shot 2018-05-23 at 6 32 28 pm

PR #1256, opened 2018-05-25
APPINT-622: Added newrelic code for updated deployment tags

pketienne:
This PR adds code which allows the newrelic gem to automatically tag deployments within the newrelic rpm portal.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require 'new_relic/recipes'
-after 'deploy:updated', 'newrelic:notice_deployment'
 
 # lock '3.6'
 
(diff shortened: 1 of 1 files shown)


carolyncole:
@informaticianme have you tested this out on QA?

config/deploy.rb
```diff
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+require 'new_relic/recipes'
+after 'deploy:updated', 'newrelic:notice_deployment'
```
awead: @informaticianme I think this can move down to line 112, and be combined, like:
`after :updated, :check_configs, 'newrelic:notice_deployment'`
Otherwise, I'd place this just after 112 to keep all the after updated calls in the same place.

awead (changes requested):

awead:
@informaticianme looks good. Can you squash your commits? Did you verify this works in QA?

pketienne:
I did test this out on QA, then incorporated Adam's recommendation, however I need to adjust something else for this PR before merging. I'm closing this and will submit another PR once I've completed the adjustment.

PR #1260, opened 2018-05-30
Index the README file in Solr

awead:
Indexing the README file in Solr improves performance because we don't have to pull files from Fedora each time a show page is displayed. Furthermore, we can correct any encoding errors in the indexed copy of the README file.

We're using the `force_encoding` option to convert to UTF-8. Any errors that result are reported when the page displays.

I can confirm this fixes the issue with this README https://scholarsphere.psu.edu/concern/generic_works/qz20ss67z which is currently not displaying correctly.

### DIFF
--- a/app/indexers/work_indexer.rb
+++ b/app/indexers/work_indexer.rb
@@ -9,6 +9,7 @@ class WorkIndexer < Sufia::WorkIndexer
       solr_doc[Solrizer.solr_name('resource_type', :facetable)] = object.resource_type
       solr_doc[Solrizer.solr_name('file_format', :stored_searchable)] = representative.file_format
       solr_doc[Solrizer.solr_name('file_format', :facetable)] = representative.file_format
+      solr_doc['readme_file_ss'] = readme_content
       solr_doc[Solrizer.solr_name(:bytes, CurationConcerns::CollectionIndexer::STORED_LONG)] = object.bytes
       SolrDocumentGroomer.call(solr_doc)
     end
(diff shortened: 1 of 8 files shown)


carolyncole:
@awead force_encoding, just tells the string to be another encoding, but does not actual translate.  I'm wondering if this is what we really want?  That assumes that everything the users put into the system is really UTF8 not Ascii.

I think indexing is a good idea, but I'm not so certain about the force.

awead:
@cam156 
One of the tests uses a ASCII-8BIT string as the README:

https://github.com/psu-stewardship/scholarsphere/pull/1260/files#diff-5d7f351680fe43e2684edcee103a14a7R88

and verifies that it goes from ASCII-8BIT to UTF-8.

I've also tested it with one of the bad READMEs in production and it fixes the issue:

https://scholarsphere.psu.edu/concern/generic_works/qz20ss67z

Is there another README that we should check? I'm open to other strategies.

carolyncole:
@awead force_encoding does not actually change the string (https://docs.ruby-lang.org/en/2.1.0/Encoding.html).  What happens if you do an actual encode there?    You are just assuming the encoding is UTF8 and was set incorrectly with a force.

app/indexers/work_indexer.rb
```diff
@@ -20,10 +21,37 @@ def representative
       object.representative || NullRepresentative.new
     end
 
+    def readme_file
+      @readme_file ||= ReadmeFile.new(object.readme_file)
+    end
+
+    def readme_content
+      return unless readme_file.content?
+      readme_file.content.encode('utf-8', invalid: :replace, undef: :replace)
```
carolyncole, line 30: 👍

carolyncole (approved):

PR #1262, opened 2018-05-31
[WIP] Display all available collections in the work dropdown

mtribone:
Removes form-control classes and adds the JS for the multi-select. Override the the CSS for the form control, create a new search builder to override 100 rows returned for the select2 list and pass Collections.count + 1000, adds new initializers for override from Curation Concerns, and some tests.


mtribone:
Using Prepend instead in another PR to alleviate unexpected behavior in the tests.

PR #1264, opened 2018-05-31
Add files to fedora as external files

little9:
Test for fedora re-direct -- if they have been
added correctly, Fedora should re-direct the
request.

Redefine the predicate for `mime_type` on the
`MetadataNode` class in the Samvera stack so
it does not conflict with the Fedora mime_type,
which is used to handle external content.

Encapsulate behavior behind the environment variable
`ENV['REPOSITORY_EXTERNAL_FILES']`. If that is set
to true the new behavior will be used. If it isn't set
to true or is missing, the old behavior will be used.

If it is true the `ENV['REPOSITORY_FILESTORE_HOST']` and
`ENV['REPOSITORY_FILESTORE']` environment variables must also
be set. See the `rails_helper` for an example of how
to set the variables.


app/jobs/ingest_file_job.rb
```diff
@@ -11,34 +15,116 @@ class IngestFileJob < ActiveJob::Base
...
+      # VersionCommitter when necessary during save_characterize_and_record_committer.
+      Hydra::Works::AddFileToFileSet.call(@file_set,
+                                        local_file,
+                                        relation,
+                                        versioning: false)
+    end
+  end
+end
+
+module ActiveFedora
```
awead: Are you overriding AF here? This should go under `config/initializers`

config/initializers/custom_mime_type_predicate.rb
```diff
@@ -0,0 +1,24 @@
...
+# results in an error delegating to `nil`.
+
+# To get around this we capture the redefined method then set
+# property back to the original method. We call this original
+# version once to set mime_type then put the redefined method back
+# in place.
+
+# `Hydra::PCDM::File` includes `WithMetadata` so we need to
+# resolve it before the method redefinition happens.
+Hydra::PCDM::File
```
awead: Does this do anything? It doesn't seem to set any variable or call a method.

spec/spec_helper.rb
```diff
@@ -1,5 +1,16 @@
...
+
+test_instance_pid = fork do
+  exec "unicorn_rails -E test -p #{HTTP_PORT}"
+end
+
+at_exit do
+  Process.kill 'INT', test_instance_pid
+  Process.wait
+end
+
```
awead: can this go in a separate file under `spec/support`. It'll get loaded automatically with rspec_helper.

awead (changes requested):

config/initializers/custom_mime_type_predicate.rb
little9: @no-reply could you say a little about this? It's not being instantiated, but needs to be loaded?

spec/spec_helper.rb
little9: Ah ok!

app/jobs/ingest_file_job.rb
little9: 👍

config/initializers/custom_mime_type_predicate.rb
no-reply: `Hydra::PCDM::File` includes `WithMetadata`, and it appears that it has to be the *first* thing in the application to do so. If we touch `WithMetadata` before `Hydra::PCDM::File` does, we get various errors related to the redefinition of `.properties`. This just makes sure that `Hydra::PCDM::File` is autoloaded before we force autoload of `WithMetadata` and do our work here.

no-reply: The comment attempts to clarify this. Can we polish it up so this can be better understood by newcomers?

no-reply:
See also: https://github.com/samvera/active_fedora/pull/1315

config/initializers/custom_mime_type_predicate.rb
awead: Ok, that makes sense. Maybe edit the comment to, "Call the clas `Hydra::PCDM::File` without any method invocation because it includes `WithMetadata` and we need to resolve it before the method redefinition happens"

bess:
Looks like that coveralls number is only for the feature test suite -- is that a concern? Do you have suggestions for how to increase it? I feel like our PR is pretty well tested but we are open to suggestions.

awead:
@bess you can generally ignore the coveralls report, it's the code climate report that's important. It's statistics are for the combined feature and unit builds.

spec/rails_helper.rb
```diff
@@ -69,6 +69,11 @@ def travis?
   include RSpec::Mocks::ExampleMethods
 end
 
+# Settings for using external files
+ENV['REPOSITORY_FILESTORE'] = Rails.root.join('public', 'repository').to_s
+ENV['REPOSITORY_FILESTORE_HOST'] = 'http://localhost:4000/repository'
+# ENV['REPOSITORY_EXTERNAL_FILES'] = 'true'
+
```
awead, line 76: @little9 I made a comment here earlier, but must have forgotten to add it and it got lost...

We use the Figaro gem to manage env variables. You should add thse variables to `config/sample/application.yml`.
Then `cp config/sample/application.yml config/application.yml` to get them locally. That will allow your test suite to run locally. It also happens in the travis script as well.

awead (changes requested):
One other small change. Otherwise, this looks good to go.

spec/rails_helper.rb
little9: Thanks! We'll add this

app/jobs/ingest_file_job.rb
```diff
@@ -11,34 +15,89 @@ class IngestFileJob < ActiveJob::Base
...
+    @file_set = file_set
+    @filepath = filepath
+    @characterize_path = filepath
     relation = opts.fetch(:relation, :original_file).to_sym
+    local_file = {}
+    local_file[:original_name] = File.basename(@filepath)
+
+    # ENV['REPOSITORY_FILESTORE'] = "/data"
+    # 1. Ensure REPOSITORY_FILESTORE exists and is writable
+    ensure_repository_filestore_exists
```
atz, line 27: This seems like a bunch of stuff better suited to a `before_perform` block, i.e. preconditions.

app/jobs/ingest_file_job.rb
```diff
@@ -11,34 +15,89 @@ class IngestFileJob < ActiveJob::Base
...
-    Hydra::Works::AddFileToFileSet.call(file_set,
-                                        local_file,
-                                        relation,
-                                        versioning: false)
+    add_file(relation, local_file)
 
     # Ensure the file set's title and label are the same as the filename
-    file_set.title = [local_file.original_name]
-    file_set.label = local_file.original_name
+    @file_set.title = [local_file.original_name]
```
atz, line 39: These `@` symbols here (and in subsequent methods) seem out of place.  You declared the `attr_reader`, so the most conventional invocation would be to use it (especially below).

PR #1271, opened 2018-06-05
Add files to fedora as external files

little9:
Test for fedora re-direct -- if they have been
added correctly, Fedora should re-direct the
request.

Redefine the predicate for `mime_type` on the
`MetadataNode` class in the Samvera stack so
it does not conflict with the Fedora mime_type,
which is used to handle external content.

Encapsulate behavior behind the environment variable
`ENV['REPOSITORY_EXTERNAL_FILES']`. If that is set
to true the new behavior will be used. If it isn't set
to true or is missing, the old behavior will be used.

If it is true the `ENV['REPOSITORY_FILESTORE_HOST']` and
`ENV['REPOSITORY_FILESTORE']` environment variables must also
be set. See the `rails_helper` for an example of how
to set the variables.

This also overrides `ActiveFedora::File`, `Hydra::Core::DownloadBehaviour`
and `CurationConcerns::WorkingDirectory` to get download functionality
working correctly.


app/jobs/ingest_file_job.rb
```diff
@@ -11,34 +15,89 @@ class IngestFileJob < ActiveJob::Base
...
+    @characterize_path = filepath
     relation = opts.fetch(:relation, :original_file).to_sym
+    local_file = {}
+    local_file[:original_name] = File.basename(@filepath)
+
+    # ENV['REPOSITORY_FILESTORE'] = "/data"
+    # 1. Ensure REPOSITORY_FILESTORE exists and is writable
+    ensure_repository_filestore_exists
+    # 2. Copy filepath to REPOSITORY_FILESTORE using a pairtree
+    copy_object_to_repository_filestore
```
bess: We should check for REPOSITORY_EXTERNAL_FILES and only write the file if it's true.

app/jobs/ingest_file_job.rb
```diff
@@ -11,34 +15,89 @@ class IngestFileJob < ActiveJob::Base
...
-    file_set.label = local_file.original_name
+    @file_set.title = [local_file.original_name]
+    @file_set.label = local_file.original_name
 
     # Persist changes to the file_set
-    file_set.save!
+    @file_set.save!
 
-    repository_file = file_set.send(relation)
+    repository_file = @file_set.files.first
```
bess: Will we ever have more than one file attached to a file set?

app/jobs/ingest_file_job.rb
```diff
@@ -11,34 +15,89 @@ class IngestFileJob < ActiveJob::Base
...
+  end
+
+  def ensure_repository_filestore_exists
+    FileUtils::mkdir_p ENV['REPOSITORY_FILESTORE']
+  end
+
+  # Given an ActiveFedora object, generate a pair tree
+  # path for it based on its id
+  # @param [ActiveFedora::Base] object
+  def self.pairtree_path(object)
```
awead: I think all the class methods should really be private instance methods. I'm not seeing any of them called from outside the class.

app/jobs/ingest_file_job.rb
```diff
@@ -11,34 +15,89 @@ class IngestFileJob < ActiveJob::Base
...
+    identifier = object.id
+    "/#{identifier[0, 2]}/#{identifier[2, 2]}/#{identifier[4, 2]}/#{identifier[6, 2]}/#{identifier}"
+  end
+
+  def self.object_path(object)
```
awead: Make this a private instance method.

app/jobs/ingest_file_job.rb
```diff
@@ -11,34 +15,89 @@ class IngestFileJob < ActiveJob::Base
...
+  def ensure_object_directory_exists(object)
+    FileUtils::mkdir_p IngestFileJob.object_path(object)
+  end
+
+  def copy_object_to_repository_filestore
+    ensure_object_directory_exists(@file_set)
+    FileUtils::cp @filepath, IngestFileJob.object_path(@file_set)
+  end
+
+  def self.pairtree_http_path(object, filepath)
```
awead: This should be a private instance method as well.

app/jobs/ingest_file_job.rb
```diff
@@ -11,34 +15,89 @@ class IngestFileJob < ActiveJob::Base
...
+    ENV['REPOSITORY_FILESTORE_HOST'] + IngestFileJob.pairtree_path(object) + '/' + File.basename(filepath)
+  end
+
+  def add_file(relation, local_file)
+    if ENV['REPOSITORY_EXTERNAL_FILES'] == 'true'
+      # Tell AddFileToFileSet service to skip versioning because versions will be minted by
+      # VersionCommitter when necessary during save_characterize_and_record_committer.
```
awead: Move this comment above the method defining and delete the other comment because they're the same.

app/jobs/ingest_file_job.rb
```diff
@@ -11,34 +15,89 @@ class IngestFileJob < ActiveJob::Base
...
     # Do post file ingest actions
     CurationConcerns::VersioningService.create(repository_file, user)
 
     # TODO: this is a problem, the file may not be available at this path on another machine.
     # It may be local, or it may be in s3
-    CharacterizeJob.perform_later(file_set, repository_file.id, filepath)
+
+    CharacterizeJob.perform_later(@file_set, repository_file.id, @characterize_path)
+  end
+
```
awead: Make all methods after `perform` private. You should be able to remove the tests for them if the lines are getting covered by other tests.

config/initializers/download_behavior.rb
```diff
@@ -0,0 +1,150 @@
+# frozen_string_literal: true
+
```
awead: @little9 You're overriding too much here. We already have a DownloadsController at that includes the DownloadBehavior module. If you need to check you can use ancestors:

```
> DownloadsController.ancestors
=> [DownloadsController, CurationConcerns::DownloadBehavior, Hydra::Controller::DownloadBehavior, ...]
```

What I would do here is create a new module that encapsulates only the changes you need to make to the module.
Something like:

``` ruby
# This module overrides certain methods in Hydra::Controller::DownloadBehavior to enable downloading of
# external content in Fedora.
module ExternalDownloadBehaviors

  def show
    if file.new_record?
      render_404
    else
      send_data file.content.read, type: file.mime_type, filename: file_name
    end
  end

  # and whatever additional methods need to change from the module
end
```

Then include this module in DownloadsController:

``` ruby
class DownloadsController < ApplicationController
  include CurationConcerns::DownloadBehavior
  include ExternalDownloadBehaviors

  [...]
```

Including the module after CurationConcerns::DownloadBehavior, which itself includes Hydra::Controller::DownloadBehavior,
will override the methods in DownloadBehavior.

config/initializers/working_directory.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
```
awead: @little9 Instead of overriding the entire service, use a prepend. This is similar to the strategy above with
DownloadBehavior, but because we don't a defining class with included modules that define behaviors, we can
use the prepend function to inject overrides into an existing class.

For example, define a module that has on the changes you want to make to WorkingDirectory, and put this
under `app/prepends/prepended_services`:

``` ruby
# This module overrides CurationConcerns::WorkingDirectory to provide support for external files.
module PrependedServices::WithExternalFileSupport

  def copy_file_to_working_directory(file, id)
    file_name = file.respond_to?(:original_filename) ? file.original_filename.match(/[^\/]+$/)[0] : ::File.basename(file)
    copy_stream_to_working_directory(id, file_name, file)
  end

  def copy_repository_resource_to_working_directory(file, id)
    Rails.logger.debug "Loading #{file.original_name} (#{file.id}) from the repository to the working directory"
    # TODO: this causes a load into memory, which we'd like to avoid
    copy_stream_to_working_directory(id, file.original_name, file.content)
  end

  private

    def copy_stream_to_working_directory(id, name, stream)
      working_path = full_filename(id, name)
      Rails.logger.debug "Writing #{name} to the working directory at #{working_path}"
      FileUtils.mkdir_p(File.dirname(working_path))
      stream.rewind
      IO.copy_stream(stream, working_path)
      working_path
    end

end
```

Note that the above module only includes the methods that you've changed in CurationConcerns::WorkingDirectory.

Then, prepend the existing CurationConcerns::WorkingDirectory class with your module in config/application.rb:

```
  CurationConcerns::WorkingDirectory.prepend PrependedServices::WithExternalFileSupport
```

awead (changes requested):

app/jobs/ingest_file_job.rb
awead: @bess this is a good question. Is `files.first` always the external file? Shouldn't it be `original_file`? There is `extracted_content` which is text content extracted from pdfs that's also stored in Fedora.

config/initializers/download_behavior.rb
little9: Thanks for this! It was extremely helpful

app/controllers/downloads_controller.rb
```diff
@@ -2,6 +2,9 @@
 
 class DownloadsController < ApplicationController
   include CurationConcerns::DownloadBehavior
+  if ENV['REPOSITORY_EXTERNAL_FILES'] == 'true'
```
awead, line 5: @little9 Have you tried this on the server with it set to both true and false? We're messing with loading, which can get dicey.

awead (changes requested):

awead (approved):

PR #1277, opened 2018-06-07
Create embargoes and leases on file sets when creating new works

### LINKED ISSUE #1050: Embargo and Lease do not show properly on the FileSet edit
This work is embargoed:
screen shot 2017-10-23 at 10 16 24 am

The FileSet does not show the embargo during edit.
screen shot 2017-10-23 at 10 14 35 am

awead:
Copies additional attributes down to the processes that are creating file sets for for new works.

Fixes #1050


carolyncole (commented):
@awead the code looks reasonable.  Does deletion test also fail on your PC? You included a fair amount of code from the parent that could have some bug.

awead:
@cam156 No, it passes locally. However, I'm also seeing the same failure in one of the other PRs, so it's possible this is some kind of transient error that was introduced with a prior PR.

carolyncole (approved):

Issue #1278, opened 2018-06-08
When using external files, newly uploaded versions have Rackmultipart in name


little9:
This is caused by the version upload form using a different mechanism than the initial upload form:
* Create File - works
* Create second Version - works, but uses Rackmultipart for file name (the tmp/ file) in the /repo directory
** the label is correct
* Restore first version - works, but changes the file label to http url
* Restore second version - works, but changes the file name to http url

PR #1279, opened 2018-06-08
Preserve the filename if given to IngestFileJob via opts

little9:
This fixes an issue where files that are uploaded via
the versions uploader retain their temporary filename.

The inital file upload form creates a temporary file with
the original filename, but the version upload does not.

Related to #1278


app/jobs/ingest_file_job.rb
```diff
@@ -22,14 +22,18 @@ def perform(file_set, filepath, user, opts = {})
     local_file = {}
     local_file[:original_name] = File.basename(@filepath)
 
+    # Whenever you upload a version, the filename option is set
```
carolyncole: Should this code be moved into the if since it is not used?

app/jobs/ingest_file_job.rb
```diff
@@ -22,14 +22,18 @@ def perform(file_set, filepath, user, opts = {})
...
+    local_file_name = opts.fetch(:filename, nil) || local_file[:original_name]
+
     if ENV['REPOSITORY_EXTERNAL_FILES'] == 'true'
-      Scholarsphere::Pairtree.new(@file_set).create_repository_files(@filepath)
+      @filepath = Scholarsphere::Pairtree.new(@file_set).create_repository_files(@filepath, local_file_name)
     end
 
     # Wrap in an IO decorator to attach passed-in options
-    local_file = Hydra::Derivatives::IoDecorator.new(File.open(filepath, 'rb'))
+    local_file = Hydra::Derivatives::IoDecorator.new(File.open(@filepath, 'rb'))
```
carolyncole: I do not believe this change is necessary since filepath is just a attr_reader

carolyncole (changes requested):
Just a couple small changes.  Looks good overall

app/jobs/ingest_file_job.rb
little9: 👍

little9:
@cam156 thanks for the feedback! I went ahead and made the changes you suggested.

carolyncole:
@little9 Unfortunately this PR conflicts with the one I just merged

little9:
@cam156 This is back to passing after resolving the conflicts.

awead (dismissed):

PR #1280, opened 2018-06-08
Add a rake task to run thge ExternalFilesConversion class

little9:
This commit introduces a rake task that creates a new versions of
a file in order to create a copy of the file that exists on external
storage.

To run the rake task:

`rake scholarsphere:internal_to_external_files`

It will iterate over all the `GenericWork`s in the repository.

### DIFF
--- a/lib/tasks/scholarsphere.rake
+++ b/lib/tasks/scholarsphere.rake
@@ -221,4 +221,17 @@ namespace :scholarsphere do
     result = work.save
     puts "Work #{work_id} marked as private: #{result}"
   end
+
+  desc 'Convert files stored internally in Fedora to externally on the filesystem'
+  task 'internal_to_external_files' => :environment do
+    puts 'This task will create a new version of all the works.'
+    puts "The process will place the content in the #{ENV['REPOSITORY_FILESTORE']} directory."
+    puts 'Are you sure you want to do this? (y/n)'
+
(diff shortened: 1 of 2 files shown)


lib/tasks/scholarsphere.rake
```diff
@@ -221,4 +221,17 @@ namespace :scholarsphere do
...
     puts "Work #{work_id} marked as private: #{result}"
   end
+
+  desc 'Convert files stored internally in Fedora to externally on the filesystem'
+  task 'internal_to_external_files' => :environment do
+    puts 'This task will create a new version of all the works.'
+    puts "The process will place the content in the #{ENV['REPOSITORY_FILESTORE']} directory."
+    puts 'Are you sure you want to do this? (y/n)'
+
+    if STDIN.gets.strip == 'y'
```
bess, line 231: I think we should put in a check here for whether this is already an external file store object.

little9: 👍

little9: @bess I added this check to the `ExternalFilesConversion` class here: https://github.com/psu-stewardship/scholarsphere/pull/1280/files#diff-ae49bb34091256892a88d5b57a89d0c2

carolyncole:
@little9 and @bess I'm going to leave this PR unreviewed since we decided against creating an additional version.

PR #1281, opened 2018-06-08
Store files in bags on the filesystem

little9:
This commit changes the `Pairtree` class so
that it stores files in a bag structure. The structure looks like:

```
/repository/ms/t3/ka/bc/mst3kabc
1528492618/
├── bag-info.txt
├── bagit.txt
├── data
│   └── world.png
├── manifest-md5.txt
├── manifest-sha1.txt
├── tagmanifest-md5.txt
└── tagmanifest-sha1.txt                                                            
```

So you have as root repistory folder, a pairtree for the idenitifier,
a unix timestamp for versioning, which is a bag.

This change uses the ruby bagit library to create the bag:
https://github.com/tipr/bagit

### DIFF
--- a/lib/scholarsphere/pairtree.rb
+++ b/lib/scholarsphere/pairtree.rb
@@ -3,8 +3,9 @@
 module Scholarsphere
   class Pairtree
     attr_reader :object
-    def initialize(object)
+    def initialize(object, bagger)
       @object = object
+      @bagger = bagger
       ensure_repository_filestore_exists
     end
 
@@ -12,8 +13,11 @@ module Scholarsphere
     # path for it based on its id
     # @param [ActiveFedora::Base] object
     def full_path
-      identifier = @object.id
-      Dir["#{ENV['REPOSITORY_FILESTORE']}/#{identifier[0, 2]}/#{identifier[2, 2]}/#{identifier[4, 2]}/#{identifier[6, 2]}/#{identifier}/*"].sort.last
+      if File.exists?(storage_dir + '/data')
+        storage_dir + '/data'
+      else
+        storage_dir
+      end
     end
 
     def path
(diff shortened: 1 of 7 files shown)


spec/lib/scholarsphere/pairtree_spec.rb
```diff
@@ -19,7 +20,7 @@
       file_set.delete
       file_set.eradicate
       if Rails.env.test?
-        FileUtils.rm_rf(ENV['REPOSITORY_FILESTORE'])
+        # FileUtils.rm_rf(ENV['REPOSITORY_FILESTORE'])
```
carolyncole: If this is undesired then can we just remove it?

lib/scholarsphere/pairtree.rb
```diff
@@ -40,6 +44,18 @@ def ensure_object_directory_exists
     def create_repository_files(filepath)
       ensure_object_directory_exists
       FileUtils::cp filepath, full_path
```
carolyncole, line 46: Would it be better to just do a move?

carolyncole (changes requested):
The build is failing and a few comments

spec/lib/scholarsphere/pairtree_spec.rb
little9: Sorry, this is desired! I was checking the contents of that directory and forgot to un-comment it.

lib/scholarsphere/pairtree.rb
little9: If I do a move there, then in the test suite the `spec/fixtures/world.png` file will be deleted and subsequent tests will fail. I was assuming that you just have some of cron job that deletes temporary files on a regular basis on the server. Moving wouldn't be more efficient here because it just combines the copy and delete/unlink operations.

carolyncole (approved):

Issue #1282, opened 2018-06-11
Empty file when downloading a work with external files as zip 


little9:
When you click 'Download Work as Zip' with externally stored files, the resulting zip file is blank.

PR #1286, opened 2018-06-11
Fix for `WorkZipService` returning and empty file with external files

little9:
This commit fixes an issue where the `WorkZipService` returns a zip
file without any content when in external files mode. It modifies
the `WorkZipService` so that it can handle internally and externally stored
files.

Related to issue #1282

### DIFF
--- a/app/services/work_zip_service.rb
+++ b/app/services/work_zip_service.rb
@@ -45,14 +45,40 @@ class WorkZipService
   private
 
     def add_files_to_zip(zipfile, file_set)
-      return unless file_set.original_file.present? && (ability.can? :read, file_set.id)
-
-      zipfile.get_output_stream(file_set.title.first) do |outfile|
-        file_set.original_file.stream.each { |buffer| outfile.write(buffer) }
+      if ENV['REPOSITORY_EXTERNAL_FILES'] == 'true'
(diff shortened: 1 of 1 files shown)


app/services/work_zip_service.rb
```diff
@@ -45,14 +45,31 @@ def call
...
+        add_internal_files_to_zip(zipfile, file_set)
+      end
+    end
+
+    def zip_manifest
+      resource.file_sets
+    end
 
+    def external_file_set_steam(file_set)
+      open(file_set.original_file.uri).read
```
carolyncole: @little9 I think you need the read ability check here see line 64

carolyncole (changes requested):
The build is failing, but I think it is becuase the check for read ability is no longer being done.

app/services/work_zip_service.rb
```diff
@@ -45,14 +45,40 @@ def call
...
+    def zip_manifest
+      resource.file_sets
+    end
+
+    def external_file_set_steam(file_set)
+      open(file_set.original_file.uri).read
+    end
 
+    def add_internal_files_to_zip(zipfile, file_set)
+      return unless add_to_zip?
```
carolyncole, line 64: 👍

carolyncole (approved):

PR #1292, opened 2018-06-13
Change all internal files versions to external files

little9:
This commit includes a class that takes a Work type and
converts each of the objects to use external storage.

It does this by going through each of the versions for all of the files.
It deletes the first version, while preserving the content. The
subsequent versions have the correct version names: 'version1', 'version2', etc.

There is a rake task to run this conversion that you can run:
`rake scholarsphere:internal_to_external_files`

### DIFF
--- a/app/services/external_files_conversion.rb
+++ b/app/services/external_files_conversion.rb
@@ -1,29 +1,63 @@
 # frozen_string_literal: true
 
-class ExternalFilesConversion
-  attr_reader :work
+require 'fileutils'
 
-  def initialize(work)
-    @work = work
+class ExternalFilesConversion
+  attr_reader :work_class, :user
+
+  def initialize(work_class)
+    @work_class = work_class
+    @user = User.batch_user
   end
 
   def convert
-    @work.all.each do |work|
-      begin
-        if work.file_sets.first.original_file.nil? == false
-          IngestFileJob.perform_now(work.file_sets.first, working_file(work.file_sets.first.original_file, work.file_sets.first.id), User.batch_user, filename: work.file_sets.first.original_file.original_name)
+    @work_class.all.each do |work|
+      work.file_sets.each do |file_set|
+        file_set.files.each do |file|
+          convert_file(work, file_set, file)
         end
-      rescue OpenURI::HTTPError
-        Rails.logger.warn "Problem accessing the URI for GenericWork: #{work.id}"
-      rescue Ldp::HttpError
-        Rails.logger.warn "Problem accessing the Fedora URI for GenericWork #{work.id}"
       end
     end
   end
 
   private
 
-    def working_file(file, id)
-      CurationConcerns::WorkingDirectory.copy_repository_resource_to_working_directory(file, id)
+    def convert_file(work, file_set, file)
+      # This slug must be prefixed with auto_ so that it will not appear in versions.all
+      ActiveFedora.fedora.connection.post(file.uri + '/fcr:versions', nil, slug: 'auto_placeholder')
+      version_contents = []
+      file.versions.all.each do |version|
+        version_content = write_version_content(version.uri)
+        version_contents << version_content
+        ActiveFedora.fedora.connection.delete(version.uri)
+      end
+
+      version_contents.each do |version_content|
+        file_set.reload
+        if work.file_sets.first.original_file.nil? == false
+          IngestFileJob.perform_now(file_set, version_content, @user)
+        end
+      end
+      ActiveFedora.fedora.connection.delete(file.uri + '/fcr:versions/auto_placeholder')
+    rescue OpenURI::HTTPError
+      Rails.logger.warn "Problem accessing the URI for GenericWork: #{work.id}"
+    rescue Ldp::HttpError
+      Rails.logger.warn "Problem accessing the Fedora URI for GenericWork #{work.id}"
+    end
+
+    def write_version_content(version_uri)
+      version_file_name = filename_from_content_disposition(open(version_uri).meta['content-disposition'])
+      time_stamp = Time.now.to_i.to_s
+      FileUtils.mkdir_p(Rails.root.join('tmp', 'external_internal_conversion', time_stamp))
+
+      file = File.new(Rails.root.join('tmp', 'external_internal_conversion', time_stamp, version_file_name), 'wb+')
+      file.write open(version_uri).read
+      file_path = File.absolute_path(file.path)
+      file.close
+      file_path
+    end
+
+    def filename_from_content_disposition(content_disposition)
+      content_disposition.split(';')[1].split('filename=')[1].split('"')[1]
     end
 end
--- a/lib/tasks/scholarsphere.rake
+++ b/lib/tasks/scholarsphere.rake
@@ -221,4 +221,22 @@ namespace :scholarsphere do
     result = work.save
     puts "Work #{work_id} marked as private: #{result}"
   end
+
+  desc 'Convert files stored internally in Fedora to externally on the filesystem'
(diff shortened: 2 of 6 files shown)


app/services/external_files_conversion.rb
```diff
@@ -11,7 +13,11 @@ def convert
     @work.all.each do |work|
       begin
         if work.file_sets.first.original_file.nil? == false
-          IngestFileJob.perform_now(work.file_sets.first, working_file(work.file_sets.first.original_file, work.file_sets.first.id), User.batch_user, filename: work.file_sets.first.original_file.original_name)
+          @file_set = work.file_sets.first
```
bess: I think this is only going to convert the first `file_set` in each work.

bess (changes requested):

spec/factory_helpers.rb
```diff
@@ -0,0 +1,67 @@
+module FactoryHelpers
```
awead: there's already a FactoryHelpers module at spec/support/helpers/factory_helper.rb. Could you add this there?

app/services/external_files_conversion.rb
```diff
@@ -26,4 +32,36 @@ def convert
...
+    end
+
+    def filename_from_content_disposition(content_disposition)
+      content_disposition.split(';')[1].split('filename=')[1].split('"')[1]
+    end
+
+    def write_version_content(version_uri)
+      version_file_name = filename_from_content_disposition(open(version_uri).meta['content-disposition'])
+
+      FileUtils.mkdir_p(Rails.root.join('tmp', 'external_internal_conversion'))
```
awead: Can this be configurable in ENV? It can default to `tmp/external_internal_conversion`

app/services/external_files_conversion.rb
```diff
@@ -26,4 +32,36 @@ def convert
...
+
+      file = File.new(Rails.root.join('tmp', 'external_internal_conversion', version_file_name), 'wb+')
+      file.write open(version_uri).read
+      file_path = File.absolute_path(file.path)
```
awead: If we're copying the file and re-ingesting, do we have any concerns about catching any checksum errors? @cam156 ?

awead (changes requested):

app/services/external_files_conversion.rb
```diff
@@ -1,29 +1,53 @@
 # frozen_string_literal: true
 
+require 'fileutils'
+
 class ExternalFilesConversion
-  attr_reader :work
+  attr_reader :work_type, :user
 
   def initialize(work)
```
carolyncole: This attribute should also be renames to indicate you are looking for a class

app/services/external_files_conversion.rb
```diff
@@ -1,29 +1,53 @@
...
 
   def convert
-    @work.all.each do |work|
-      begin
-        if work.file_sets.first.original_file.nil? == false
-          IngestFileJob.perform_now(work.file_sets.first, working_file(work.file_sets.first.original_file, work.file_sets.first.id), User.batch_user, filename: work.file_sets.first.original_file.original_name)
+    @work_type.all.each do |work|
+      work.file_sets.each do |file_set|
+        file_set.files.each do |file|
+          ActiveFedora.fedora.connection.post(file.uri + '/fcr:versions', nil, slug: 'auto_placeholder')
```
carolyncole: can we add a comment here to indicate that the auto is important in the name?

app/services/external_files_conversion.rb
```diff
@@ -1,29 +1,53 @@
...
-        Rails.logger.warn "Problem accessing the Fedora URI for GenericWork #{work.id}"
       end
     end
   end
 
   private
 
-    def working_file(file, id)
-      CurationConcerns::WorkingDirectory.copy_repository_resource_to_working_directory(file, id)
+    def filename_from_content_disposition(content_disposition)
```
carolyncole: My preference for reading would be to have this method after the method that uses it.

app/services/external_files_conversion.rb
```diff
@@ -1,29 +1,53 @@
...
+      content_disposition.split(';')[1].split('filename=')[1].split('"')[1]
+    end
+
+    def write_version_content(version_uri)
+      # TO-DO: Use the version label in folder for writing content to keep files w/ same name
+      version_file_name = filename_from_content_disposition(open(version_uri).meta['content-disposition'])
+      time_stamp = Time.now.to_i.to_s
+      FileUtils.mkdir_p(Rails.root.join('tmp', 'external_internal_conversion', time_stamp))
```
carolyncole: Is this going to make a different directory for each temporary file?

lib/tasks/scholarsphere.rake
```diff
@@ -221,4 +221,22 @@ namespace :scholarsphere do
...
   end
+
+  desc 'Convert files stored internally in Fedora to externally on the filesystem'
+  task 'internal_to_external_files' => :environment do
+    unless ENV['REPOSITORY_EXTERNAL_FILES'] == 'true'
+      puts 'You must set the REPOSITORY_EXTERNAL_FILES environment variable to true'
+      puts 'before running this task'
+      exit
+    end
+    puts 'This task will create a new version of all the works.'
```
carolyncole, line 232: 👍

spec/support/helpers/factory_helpers.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 module FactoryHelpers
+  extend self
```
carolyncole, line 4: What does this do?

carolyncole (commented):
I'm assuming this is still a work in progress, but I added some comments.

carolyncole:
Also could you rebase against develop?  @awead added another build so travis checks both with and without external files.

app/services/external_files_conversion.rb
little9: Yeah, this is so that names don't collide

app/services/external_files_conversion.rb
little9: 👍

app/services/external_files_conversion.rb
little9: 👍

bess (approved):
:+1:

spec/support/helpers/factory_helpers.rb
little9: It makes all of the methods instance methods for the class the includes the module

app/services/external_files_conversion.rb
```diff
@@ -1,29 +1,54 @@
...
+    @work_class = work_class
+    @user = User.batch_user
   end
 
   def convert
-    @work.all.each do |work|
-      begin
-        if work.file_sets.first.original_file.nil? == false
-          IngestFileJob.perform_now(work.file_sets.first, working_file(work.file_sets.first.original_file, work.file_sets.first.id), User.batch_user, filename: work.file_sets.first.original_file.original_name)
+    @work_class.all.each do |work|
```
carolyncole, line 14: Can we check to see if the file is external and not do a conversion if it is already external.

app/services/external_files_conversion.rb
```diff
@@ -1,29 +1,54 @@
...
+          file.versions.all.each do |version|
+            version_content = write_version_content(version.uri)
+            version_contents << version_content
+            ActiveFedora.fedora.connection.delete(version.uri)
+          end
+
+          version_contents.each do |version_content|
+            file_set.reload
+            IngestFileJob.perform_now(file_set, version_content, @user)
+          end
```
carolyncole: Rescue any exception and put out an error log, and a way to restart

carolyncole (approved):

PR #1299, opened 2018-06-14
Remove the Blocks with N/A from the Admin Stats

### LINKED ISSUE #562: Admin page show n/a for stats
When I click on Administration
Then I get forwarded to the Administration page
Then I see n/a for total visitors, returning visitors, and new visitors

screen shot 2017-01-19 at 7 19 32 am

mtribone:
Overrides the Sufia Admin Show partial to remove the blocks without any data. Fixes #562 

screen shot 2018-06-14 at 6 33 12 pm

### DIFF
--- a/app/views/sufia/admin/show.html.erb
+++ b/app/views/sufia/admin/show.html.erb
@@ -1,0 +1,13 @@
+<%# Removing the columns for total, returning, and new visitors because we do not have data for them %>
+<div class="row">
+  <div class="col-md-3">
+    <div class="panel panel-default">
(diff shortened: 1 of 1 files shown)


app/views/sufia/admin/show.html.erb
```diff
@@ -0,0 +1,12 @@
+<div class="row">
```
carolyncole, line 2: @mtribone can you add a comment to the top of this to indicate why we are overriding it.

carolyncole (changes requested):

carolyncole (approved):

Issue #978, opened 2017-07-07
Proxy depositor is not getting set correctly


awead:
I can upload a new work on behalf of @cam156 and it looks like:
```
> work.on_behalf_of
=> "cam156"
> work.depositor
=> "cam156"
> work.proxy_depositor
=> "cam156"
```
But I would expect `proxy_depositor` to be me and not her.

carolyncole:
@awead was this work completed?

awead:
@cam156 no, I don't believe so, but we should verify it's still a problem.

awead:
@cam156 Behavior is still the same. Do you think this is correct? @olendorf

olendorf:
@awead  its not correct, it should put the "blame" on the proxy_depositor correctly. It is actually important for matters of provenance and trust too.

awead:
I think the problem was introduced in #895, which was meant to fix another problem. Given the race condition, I'm not sure of a clear way to fix this. Perhaps using `current_user` as the value for proxy depositor.

PR #1337, opened 2018-08-16
Enable editing of on_behalf_of

### LINKED ISSUE #979: Administrators Edit Proxy deposit information
Allow Administrators access to proxy deposit information for editing.

Allow an administrator to change the depositor and the on behalf of fields when editing.

awead:
Allows and administrator to change the on_behalf_of user for works that have either been transferred or uploaded by a proxy user.

Fixes #979

### DIFF
--- a/app/actors/curation_concerns/actors/generic_work_actor.rb
+++ b/app/actors/curation_concerns/actors/generic_work_actor.rb
@@ -29,9 +29,10 @@ module CurationConcerns
         # if the user is depositing on behalf of someone else.
         def apply_save_data_to_curation_concern(attributes)
           if attributes.fetch('on_behalf_of', nil).present?
-            depositor = ::User.find_by_user_key(attributes.fetch('on_behalf_of'))
-            curation_concern.apply_depositor_metadata(depositor)
-            curation_concern.edit_users += [depositor, user.user_key]
+            current_depositor = curation_concern.depositor
+            new_depositor = ::User.find_by_user_key(attributes.fetch('on_behalf_of'))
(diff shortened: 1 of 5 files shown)


app/actors/curation_concerns/actors/generic_work_actor.rb
```diff
@@ -31,7 +31,14 @@ def apply_save_data_to_curation_concern(attributes)
           if attributes.fetch('on_behalf_of', nil).present?
             depositor = ::User.find_by_user_key(attributes.fetch('on_behalf_of'))
             curation_concern.apply_depositor_metadata(depositor)
-            curation_concern.edit_users += [depositor, user.user_key]
+
+            if user.administrator?
+              curation_concern.edit_users = [depositor, curation_concern.proxy_depositor]
```
carolyncole: @awead would curation_concern.proxy_depositor work for both the admin and the regular user case?

carolyncole (approved):
@awead my comment is more of a question, not a necessary change.

app/actors/curation_concerns/actors/generic_work_actor.rb
awead: @cam156 I think so, yes. There's some other bug going on here, too, that's related to #978. Also, the problem with the above line is that would remove any others' permissions if they had been added separately from the proxy or on-behalf-of user, so I think I need to correct for that.

awead:
@cam156 I pushed up a new change that should address your previous comment, and I believe also address the issue of any pre-existing edit users who aren't either depositor or proxy.

Issue #1338, opened 2018-08-16
External files migration fails due to encoding errors


awead:
When migrating the binaries out of Fedora, content that has incorrect UTF8 encodings causes the whole process to fail.

We should be able to use EncodingService to correct the issue.

Child of #1128

PR #1339, opened 2018-08-16
Convert a single object

bess:
If we can convert a single object from internal to
external files storage, instead of having to convert
all of the objects of a given class, it will make it
much faster to debug configuration issues.

Connected to #1338

### DIFF
--- a/app/services/external_files_conversion.rb
+++ b/app/services/external_files_conversion.rb
@@ -3,24 +3,40 @@
 require 'fileutils'
 
 class ExternalFilesConversion
-  attr_reader :work_class, :user
+  attr_reader :work_class, :user, :opts
 
-  def initialize(work_class)
+  def initialize(work_class, opts = {})
     @work_class = work_class
     @user = User.batch_user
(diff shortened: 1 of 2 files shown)


app/services/external_files_conversion.rb
```diff
@@ -3,24 +3,40 @@
...
 
 class ExternalFilesConversion
-  attr_reader :work_class, :user
+  attr_reader :work_class, :user, :opts
 
-  def initialize(work_class)
+  def initialize(work_class, opts = {})
     @work_class = work_class
     @user = User.batch_user
+    @opts = opts
```
awead, line 11: @bess where does `opts` get used?

awead (changes requested):

app/services/external_files_conversion.rb
bess: Whoops, that was an approach I started to take but then decided to go a different way. I'll remove it.

bess:
Reopening elsewhere

PR #1341, opened 2018-08-17
Replace system create date with date uploaded

### LINKED ISSUE #933: Stats report is using system created date instead of migrated created date
The stats report is not looking at our migrated field, but must be looking at the actual system creation date.

This is ok, except when we migrate data.  This morning we got a report that 7324 items were uploaded last week.

awead:
Uses the date uploaded field in Solr to query for works in the stats reports.

Fixes #933

### DIFF
--- a/app/prepends/prepended_services/with_date_uploaded.rb
+++ b/app/prepends/prepended_services/with_date_uploaded.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+# Replaces the system create date with the uploaded data in Sufia::QueryService
+# so that reports are generated based on the date a work was originally deposited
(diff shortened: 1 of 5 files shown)


awead:
This could use some kind of "feature" test, but I'm unsure of where that ought to go.

carolyncole (commented):
@awead I agree it would be nice to have a feature test to make sure this will work with migrated data on the Stats page.  Is the stats page working at this point?

awead:
@cam156 the stats page isn't working, presently, but we do get the report emailed to us, so perhaps I could do a feature test there.

carolyncole (approved):

Issue #1175, opened 2018-02-13
The admin/stats url on the qa environment is returning a ss error page


pketienne:
I'm working specifically with the QA environment but I have successfully and completely synced the stage environment with qa so they should be identical at this point (but are not with respect to the `/admin/stats` page).

In the original issue, the error was `undefined method 'name' for nil:NilClass` (#1091), however the issue I'm getting now is `undefined method '[]' for nil:NilClass`.

```
Rendering 500 page due to exception: #<NoMethodError: undefined method `[]' for nil:NilClass> - ["/opt/heracles/deploy/scholarsphere/shared/bundle/ruby/2.3.0/gems/sufia-7.4.      1/app/services/sufia/statistics/term_query.rb:26:in `query'", "/opt/heracles/deploy/scholarsphere/shared/bundle/ruby/2.3.0/gems/sufia-7.4.1/app/presenters/sufia/admin_stats_presenter.rb:31:in `top_formats'",       "/opt/heracles/deploy/scholarsphere/shared/bundle/ruby/2.3.0/gems/sufia-7.4.1/app/views/sufia/admin/stats/_top_data.html.erb:9: [...]
```
I've synced the following persistent data sources:
* solr
* fedora
* redis
* mariadb - local - fedora
* mariadb - remote - rails
* uploads folder (derivatives)

I've also run the following code on the QA environment (after sync with stage):
```ruby
depositors = Sufia::Statistics::Depositors::Summary.new(nil,nil).depositors
depositors.each{|dep| dep[:user].name rescue User.create!(login:dep[:key])}
```

Both commands seem to have completed successfully with output related to users.

The problem stems from the `term_query.rb` file, line 26. It seems that either `json['terms']` or `json['terms'][term]` is an empty reference, but after looking through related files in the stacktrace, I've not been able to determine why this might be the case.

Any tips, hints, or suggestions are appreciated.

awead:
Terms query needs to be enable in solrconfig. See #1346.

The no method error was being raised because the json response from the term query was returning `{}` and calling `[term]` caused the error. Oddly, RSolr doesn't error when we call a term query that's not configured, but the error does appear in Travis:
```
<p>
    Showing <i>/home/travis/build/psu-stewardship/scholarsphere/vendor/bundle/ruby/2.3.0/gems/sufia-7.4.1/app/views/sufia/admin/stats/_top_data.html.erb</i> where line <b>#9</b> raised:
  </p>
[...]

PR #1353, opened 2018-08-22
Better processing of term queries in Solr

awead:
Increases test coverage for TermQuery and handles any exception that
might be returned from RSolr.

Related to #1175


spec/services/sufia/term_query_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  after(:all) do
+    ActiveSupport::Dependencies.remove_constant('ByMyTerm')
+  end
+
+  let(:term) { ByMyTerm.new }
+
+  context 'when Solr is configured to respond' do
+    it 'returns an accurate response' do
+      expect(term.query).to eq([])
```
carolyncole: @awead since you are not getting any values here, you should also check to be certain an error message is not being logged.

app/prepends/prepended_services/with_null_term_query.rb
```diff
@@ -13,10 +13,13 @@ def query
...
                                                   omitHeader: 'true' }
     if json.blank?
-      Rails.logger.error 'Unable to reach TermsComponent via Solr connection. Is it enabled in your solr config?'
+      Rails.logger.error "Solr returned an empty response for term query #{term}. Is it configured correctly?"
       return []
     end
 
     Sufia::Statistics::TermQuery::Result.build(json['terms'][term])
+  rescue StandardError => exception
+    Rails.logger.error "Unable query Solr for term #{term}: #{exception}"
```
carolyncole: `Unable to query Solr` might be better grammatically.

carolyncole (changes requested):

awead:
@cam156 this is fixed up.

PR #1354, opened 2018-08-22
Use EncodingService in FileSetIndexer

little9:
This commit changes the `FileSetIndexer` to use the
`EncodingService`. This will ensure that whatever has
been stored in 'all_text_timv' is converted to UTF-8.

`force_encoding` will only treat the strings stored there
as if they were UTF-8.

### DIFF
--- a/spec/indexers/file_set_indexer_spec.rb
+++ b/spec/indexers/file_set_indexer_spec.rb
@@ -42,12 +42,12 @@ describe FileSetIndexer do
         allow(file_set).to receive(:extracted_text).and_return(extracted)
       end
       it 'forces utf8' do
-        expect(text).to receive(:force_encoding).with('UTF-8').and_return('abc123')
+        expect(text).to receive(:encode).with('utf-8', invalid: :replace, undef: :replace).and_return('abc123')
         subject
       end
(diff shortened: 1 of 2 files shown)


spec/indexers/file_set_indexer_spec.rb
```diff
@@ -42,12 +42,12 @@
         allow(file_set).to receive(:extracted_text).and_return(extracted)
       end
       it 'forces utf8' do
-        expect(text).to receive(:force_encoding).with('UTF-8').and_return('abc123')
+        expect(text).to receive(:encode).with('utf-8', invalid: :replace, undef: :replace).and_return('abc123')
```
awead, line 45: @little9 I think this is fine, but the test is actually coupled too closely to the service's implementation. You might say that all the indexer is concerned with is calling the service, so simply having `expect(EncodingService).to receive(:call).with(extracted_text)` would be a better fit for the unit test.

With your test, you're actually testing what the service is doing, and technically, the indexer doesn't know or care about that. Some might argue the opposite; however, I don't think it's worth holding up anything over this.

awead (approved):

PR #1364, opened 2018-08-24
Adding tests and refactoring CopyPermissionsJob

carolyncole:
Added tests to cover embargo and lease copy

Co-authored-by: Adam Wead <amsterdamos@gmail.com>
Co-authored-by: Derek Morr <derekmorr@psu.edu>


carolyncole:
This is failing code climate, but the previous code was failing with cyclic complexity of 15 instead of 8 so this code is much better.

Issue #1369, opened 2018-08-27
Migration failing with out of memory error


bess:
When we run the external files migration on dce-fedora, it currently fails with an out of memory error

- [ ] Refactor the migration task so it works from an array of strings instead of loading all objects into memory

Child of #1128

PR #1370, opened 2018-08-27
Do not load all objects into memory during migration

bess:
In order to use memory more effectively, don't load all
objects into memory during external files migration.
Instead, work from an array of string ids.

Connected to #1369

### DIFF
--- a/app/services/external_files_conversion.rb
+++ b/app/services/external_files_conversion.rb
@@ -14,13 +14,16 @@ class ExternalFilesConversion
   end
 
   # If we receive a work ID, only convert that one item
+  # If we receive a file of IDs, convert that list
   # Otherwise, convert all instances of the given class
-  # @param [String] id
-  def convert(id = nil)
+  # @param [Hash] opts optional parameters. Valid values are :id and :file
+  def convert(opts = {})
     start_time = Time.now
     @logger.info "Starting conversion process at #{start_time}"
-    if id
-      convert_work(@work_class.find(id))
+    if opts[:id]
+      convert_work(opts[:id])
+    elsif opts[:file]
+      convert_from_file(opts[:file])
     else
       convert_class
     end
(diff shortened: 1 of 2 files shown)


spec/services/external_files_conversion_spec.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
+require 'feature_spec_helper'
```
awead: I'm surprised you needed this. What's the reason?

app/services/external_files_conversion.rb
```diff
@@ -33,22 +33,30 @@ def convert(id = nil)
   private
 
     def convert_class
-      all_objects = @work_class.all
+      all_objects = ActiveFedora::SolrService.query("has_model_ssim:#{@work_class}", fl: 'id,depositor_tesim', rows: 1000000).map(&:id)
```
awead: Why do we need a depositor? Just curious, really...

awead (commented):
I had some questions, but no objections to merging.

bess:
@awead I thought I needed the feature spec cleanup job b/c I was getting test failures b/c something wasn't getting cleaned up. My mistake, though, it isn't necessary and I've removed it.
I also removed the unneeded depositor fl param from the solr query. It was only there because I copied the code from something Carolyn gave me and I didn't look at it carefully. My bad, again, and it's fixed.

awead:
@bess no worries! Like I said, it was only questions. Removing the feature spec helper might speed things up slight on that particular test... not that we'd notice saving a few second on one test. Our whole suite needs a lot of help.

Thanks for addressing my questions.

bess:
@awead This is green apart from some CI weirdness and it sounds like you're okay with it. Would you mind officially approving it so it can get merged?

PR #1374, opened 2018-08-28
Adding external builds to Code Climate coverage

awead:
Presently, Code Climate is only using the unit and feature test builds to calculate coverage, and is not taking the tests for Fedora's external binary storage into account. This corrects the issue by providing another parameter to the named build file so that we should have 4 files used to compute coverage. For example:

* codeclimate.421242764.feature.external-true.json
* codeclimate.421242764.unit.external-true.json
* codeclimate.421242764.feature.external-false.json
* codeclimate.421242764.unit.external-false.json


awead:
No worth it. :)

PR #1379, opened 2018-08-30
When bagging file don't read the whole file into memory

little9:
This changes the bagger class so that it doesn't
read the whole file into memory when sending the
file to BagIt. Instead it sends the data a line at a
time.


lib/scholarsphere/bagger.rb
```diff
@@ -7,7 +7,7 @@ def initialize(full_path: base_path, working_file: file)
       Dir.chdir(full_path)
       @bag = BagIt::Bag.new(full_path)
       @bag.add_file(working_file) do |io|
-        io.puts File.open(working_file).read
+        IO.foreach(working_file).each { |line| io.puts line }
```
awead, line 10: @little9 Will this work with binary data?

awead (changes requested):

lib/scholarsphere/bagger.rb
little9: @awead it does work with binary data, I'll add a test to this PR to illustrate

little9: @awead This PR now includes a spec that shows that this is working as one would expect. It makes two bags, one for text data and one for bin data and compares the SHAs in the bag with the correct SHA.

awead (approved):

Issue #1340, opened 2018-08-17
Write external files migration errors to a special log


bess:
Currently, we log errors that occur during the external files migration process because we don't want to halt the migration. See, e.g., https://github.com/psu-stewardship/scholarsphere/blob/develop/app/services/external_files_conversion.rb#L42-L45

- [ ] Instead of writing these errors to the general rails log, write them to a special log file that's easier to review (and re-run?) w/in the context of running this migration

Child of #1128

Issue #1302, opened 2018-06-15
Do not Migrate external Content


carolyncole:
In  app/services/external_files_conversion.rb check before running the conversion to see if the file is already external.

child of #1128

PR #1384, opened 2018-08-31
Refactor external files migration to avoid OOM error

bess:
* Use a default value for migration log
* Split list of pids into files
* Convert all works using files
* Write error PIDs to an error file

Connected to #1369 
Connected to #1340 
Connected to #1302

### DIFF
--- a/app/services/external_files_conversion.rb
+++ b/app/services/external_files_conversion.rb
@@ -4,13 +4,20 @@ require 'fileutils'
 require 'logger'
 
 class ExternalFilesConversion
-  attr_reader :work_class, :user
+  NUMBER_OF_PIDS_PER_FILE = 1000
+  attr_reader :work_class, :user, :pid_lists, :error_file, :logger
 
   def initialize(work_class)
     @work_class = work_class
     @user = User.batch_user
-    @logger = Logger.new(ENV['REPOSITORY_MIGRATION_LOG'])
-    @logger.level = Logger::DEBUG
+    @logger = Logger.new(ENV['REPOSITORY_MIGRATION_LOG'] || Rails.root.join('log', "external_files_conversion_#{timestamp}.log").to_s)
+    logger.level = Logger::DEBUG
+    @pid_lists = []
+    @error_file = ENV['REPOSITORY_MIGRATION_ERROR_LOG'] || Rails.root.join('log', "external_files_conversion_errors_#{timestamp}.log").to_s
+  end
+
+  def timestamp
+    @timestamp ||= Time.now.strftime('%Y-%m-%e-%H-%M-%S')
   end
 
   # If we receive a work ID, only convert that one item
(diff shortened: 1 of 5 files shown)


app/services/external_files_conversion.rb
```diff
@@ -4,13 +4,16 @@
...
 class ExternalFilesConversion
-  attr_reader :work_class, :user
+  NUMBER_OF_PIDS_PER_FILE = 1000
+  attr_reader :work_class, :user, :pid_lists, :error_file
 
   def initialize(work_class)
     @work_class = work_class
     @user = User.batch_user
-    @logger = Logger.new(ENV['REPOSITORY_MIGRATION_LOG'])
+    @logger = Logger.new(ENV['REPOSITORY_MIGRATION_LOG'] || Rails.root.join('log', "external_files_conversion_#{Time.now.strftime('%Y-%m-%e-%H-%M-%S')}.log").to_s)
```
awead: You might want to memoize the named timestamp so it'll be the same and reusable for each log file and directory.
`@logger = Logger.new(ENV['REPOSITORY_MIGRATION_LOG'] || "#{timestamped_run}.log")`
```
def timestamped_run
  @timestamped_run ||= Rails.root.join('log', "external_files_conversion_#{Time.now.strftime('%Y-%m-%e-%H-%M-%S')}.log")
end

File: app/services/external_files_conversion.rb [hunk 2]
```diff
@@ -35,12 +40,36 @@ def convert(opts = {})
...
+    # return [Array]
+    def all_objects
+      @all_objects ||= ActiveFedora::SolrService.query("has_model_ssim:#{@work_class}", fl: 'id', rows: 1000000).map(&:id)
+    end
+
+    # Create lists of PIDs for conversion, so that we can break the full export
+    # down into smaller pieces
+    # @return [Array] an array of the files that were created
+    def create_lists
+      @pid_files_dir ||= Rails.root.join('tmp', "external_files_conversion", Time.now.strftime('%Y-%m-%e-%H-%M-%S'))
```
awead: I don't think this needs to be memoized if it's only used in this method. Alternatively, you could make `pid_files_dir` its own method, which memoizes and creates it if needed:
```
def pid_files_dir
  @pid_files_dir ||= begin
    FileUtils.mkdir_p timestamped_run
    timestamped_run
  end
end
```

app/services/external_files_conversion.rb
```diff
@@ -4,13 +4,16 @@
 require 'logger'
 
 class ExternalFilesConversion
-  attr_reader :work_class, :user
+  NUMBER_OF_PIDS_PER_FILE = 1000
+  attr_reader :work_class, :user, :pid_lists, :error_file
```
awead: need another attr_reader for logger

app/services/external_files_conversion.rb
```diff
@@ -35,12 +40,36 @@ def convert(opts = {})
...
+    # down into smaller pieces
+    # @return [Array] an array of the files that were created
+    def create_lists
+      @pid_files_dir ||= Rails.root.join('tmp', "external_files_conversion", Time.now.strftime('%Y-%m-%e-%H-%M-%S'))
+      FileUtils.mkdir_p @pid_files_dir
+      lists_of_pids = all_objects.each_slice(NUMBER_OF_PIDS_PER_FILE).to_a
+      lists_of_pids.each_with_index do |list, index|
+        file_path = "#{@pid_files_dir}/#{index}.txt"
+        File.open(file_path, 'w') { |file| file.puts(list) }
+        @pid_lists << file_path
```
awead: shouldn't need the @ here.

app/services/external_files_conversion.rb
```diff
@@ -35,12 +40,36 @@ def convert(opts = {})
...
+    # Do this by first creating lists of the pids to work from, then running
+    # each of these individually, so we don't run out of memory. Pause between
+    # files.
     def convert_class
-      all_objects = ActiveFedora::SolrService.query("has_model_ssim:#{@work_class}", fl: 'id', rows: 1000000).map(&:id)
-      all_objects_count = all_objects.count
-      @logger.info "Converting #{all_objects_count} objects of type #{@work_class}"
-      all_objects.each do |work_id|
-        convert_work(work_id)
+      @logger.info "Converting #{all_objects.count} objects of type #{@work_class}"
```
awead: No @ if you use the attr_reader.

app/services/external_files_conversion.rb
```diff
@@ -64,12 +93,14 @@ def convert_work(work_id)
             convert_file(work, file_set, file)
           end
         end
+      @logger.info "Finished converting work #{work_id}"
       rescue ActiveFedora::ObjectNotFoundError => error
         @logger.error "error finding object to migrate: #{work_id}; #{error}"
+        File.open(@error_file, 'a') { |e| e.puts(work_id) }
```
awead: also no @ because you have the attr_reader

app/services/external_files_conversion.rb
```diff
@@ -88,7 +119,12 @@ def convert_file(work, file_set, file)
           IngestFileJob.perform_now(file_set, version_content, @user)
         end
       end
-      ActiveFedora.fedora.connection.delete(file.uri + '/fcr:versions/auto_placeholder')
+      # Try to delete the auto_placeholder snapshot, but if it is the only version snapshot
+      # it will raise an error. Ignore these.
+      begin
+        ActiveFedora.fedora.connection.delete(file.uri + '/fcr:versions/auto_placeholder')
+      rescue
```
awead: We should rescue the specific AF error. This would rescue any StandardError, which could have unintended consequences... unless that what you want :) In which case, `rescue StandardError` to be clear.

awead (changes requested):

bess:
@awead Take another look?

app/services/external_files_conversion.rb
```diff
@@ -19,57 +26,98 @@ def initialize(work_class)
...
+    # return [Array]
+    def all_objects
+      @all_objects ||= ActiveFedora::SolrService.query("has_model_ssim:#{@work_class}", fl: 'id', rows: 1000000).map(&:id)
+    end
+
+    # Create lists of PIDs for conversion, so that we can break the full export
+    # down into smaller pieces
+    # @return [Array] an array of the files that were created
+    def create_lists
+      pid_files_dir = Rails.root.join('tmp', 'external_files_conversion', Time.now.strftime('%Y-%m-%e-%H-%M-%S'))
```
awead: @bess timestamp here too?

app/services/external_files_conversion.rb
```diff
@@ -19,57 +26,98 @@ def initialize(work_class)
...
-      all_objects = ActiveFedora::SolrService.query("has_model_ssim:#{@work_class}", fl: 'id', rows: 1000000).map(&:id)
-      all_objects_count = all_objects.count
-      @logger.info "Converting #{all_objects_count} objects of type #{@work_class}"
-      all_objects.each do |work_id|
-        convert_work(work_id)
+      logger.info "Converting #{all_objects.count} objects of type #{@work_class}"
+      create_lists
+      pid_lists.each do |list|
+        convert_from_file(list)
+        sleep 10.minutes if Rails.env.production?
```
awead, line 76: @bess wow, 10 minutes?!

awead (changes requested):
@bess just two more last questions. Thanks!

awead (approved):
I'm good with this, but @cam156 should check on the 10  minute wait.

Issue #1301, opened 2018-06-15
Difference the Fedora Check Sum on migration


carolyncole:
When we move the data from fedora to disk we should verify the checksum to be certain we have created the file correctly.

child of #1128

PR #1389, opened 2018-09-04
Add additional checksum verification in specs

little9:
This adds an additional checksum verification to
the external files conversion spec. This ensures
that the FixityService returns the sha1 we expect
for the world.png file.

It also ensures that the checksums are the same
when the existing file is placed in `/tmp` and
that the file that is downloaded from a redirect
also has the correct checksum.

Connected to #1301


awead (approved):

little9:
I'm going to close this because the other PR introduces the Checksummer class, which is tested and is used the conversion process.

PR #1391, opened 2018-09-05
Check that checksums remain the same during conversion

little9:
This adds a class that when given a work, returns an
array with all the checksums for all versions of the files
in that work. This is used in the conversion class
before the conversion and after the conversion.

If these checksums don't match, an error is logged.

### DIFF
--- a/app/services/checksummer.rb
+++ b/app/services/checksummer.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+require 'digest'
+
+class Checksummer
+  # This class returns an array of all the checksums for all the files
+  # (and their versions) in a work.
+  # @param work [ActiveFedora::Base]
+  def initialize(work)
+    @work = work
+  end
+
+  # @return [Array] an array of checksums from the FixityService
+  def fedora_checksums
+    checksums = []
+    @work.file_sets.each do |file_set|
+      file_set.files.each do |file|
+        file.versions.all.each do |version|
+          checksums << ActiveFedora::FixityService.new(version.uri).expected_message_digest.gsub('urn:sha1:', '')
+        end
+      end
+    end
+    checksums
+  end
+
+  # @return [Array] an array of checksums generated from the files on disk
+  def disk_checksums
+    checksums = []
+    @work.file_sets.each do |file_set|
+      file_set.files.each do |file|
+        file.versions.all.each do
+          sha1 = Digest::SHA1.new
+          open(file_set.original_file.uri,
+               http_basic_authentication: [ActiveFedora.fedora_config.credentials['user'],
+                                           ActiveFedora.fedora_config.credentials['password']],
+               allow_redirections: :all) { |f| f.each_line { |line| sha1.update line } }
+          checksums << sha1.hexdigest
+        end
+      end
+    end
+    checksums
+  end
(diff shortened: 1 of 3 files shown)


bess (changes requested):
Hey, Jamie. Let's talk about this. I don't think the Fedora checksum is re-calculating the checksum of the file after it's been written to disk. I'd be very happy to be wrong on that, but that was my understanding from conversations with Esmé. So I think that post-migration we have to explicitly recalculate the checksum against the file on disk.

little9:
@bess I updated this so that it compares against a re-calculated check sum.

bess (approved):
👍 🚀

carolyncole:
@little9 It looks like the tests do not pass on Travis.  I am not certain why, but the checksums are not the same.

carolyncole:
I tried this branch on QA and got this error:
```
2018-09-06 19:37:58 -0400 (31440) Starting to convert work 8s45q881f
2018-09-06 19:38:40 -0400 (31440) error migrating object: 8s45q881f; redirection forbidden: http://scholarsphere-repo-qa.libraries.psu.edu:8080/SSqaFedora/rest/prod/3x/s5/5m/95/3xs55m9505/files/5bd5b892-d65f-43b8-8355-3759469ccada/fcr:versions/version1 -> https://scholarsphere-qa.libraries.psu.edu/binaries/3x/s5/5m/95/3xs55m9505/1536277087/data/3xs55m9505_version1_Andragogy_and_Technology_pdf.pdf
```

little9:
@cam156 that looks like the same http/https issue that was happening earlier

app/services/checksummer.rb
```diff
@@ -0,0 +1,40 @@
...
+    @work = work
+  end
+
+  # @return [Array] an array of checksums from the FixityService
+  def fedora_checksums
+    checksums = []
+    @work.file_sets.each do |file_set|
+      file_set.files.each do |file|
+        file.versions.all.each do |version|
+          checksums << ActiveFedora::FixityService.new(version.uri).expected_message_digest.gsub('urn:sha1:', '')
```
carolyncole, line 19: I wonder if this fixity service needs to be updated to use the open with the allow http to https redirect

carolyncole: Just ran a quick test on QA and this line is not an issue

app/services/checksummer.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  # @return [Array] an array of checksums generated from the files on disk
+  def disk_checksums
+    checksums = []
+    @work.file_sets.each do |file_set|
+      file_set.files.each do |file|
+        file.versions.all.each do |version|
+          sha1 = Digest::SHA1.new
+          open(version.uri) { |f| f.each_line { |line| sha1.update line } }
```
carolyncole: This line is the issue.  It does just a plain open.  It really needs to be 
```
open(
        file_set.original_file.uri,
          http_basic_authentication: [ActiveFedora.fedora_config.credentials['user'], ActiveFedora.fedora_config.credentials['password']],
          allow_redirections: :all
```

little9:
@cam156 I updated the open like you suggested and pushed that changed

PR #1491, opened 2018-11-19
Use mini_racer instead of therubyracer

awead:
The mini_racer gem fulfils the same requirements as therubyracer for our needs, but is lighter-weight and easier to install.


carolyncole (commented):
@awead looks ok. My only question is what are the installation instructions?  Do we need to update the installation wiki?

awead:
@cam156 Installing is just the usual `gem` or `bundle install`. I checked the wiki page on installing from a clean Mac, and there aren't any references to `therubyracer` or its dependencies.

Issue #1237, opened 2018-04-18
Move from DOI to DataCite


carolyncole:
Switch the API from EZID to Datacite. Should be done by the end of 2018. Replace the EZID gem with the DataCite gem and make any necessary changes.

awead:
We will need to verify which gem we'll need to user. @olendorf will do this, then we can point the ticket.

carolyncole:
@olendorf do you have any idea which gem we would switch.

carolyncole:
https://support.datacite.org/docs/datacite-ez-api-guide Is this the location?

carolyncole:
Add processes for migrating the old IDs from the EZID system.

carolyncole:
We are waiting on purchasing for the real license key.  It can take about a month to get the real key, so this may need to be updated after the sprint is complete.

@olendorf  will ask for early sandbox access for testing.

awead:
Based on this link: https://support.datacite.org/docs/datacite-ez-api-guide#section-how-to-use
We can continue to use the same ezid-client gem we've been using. Is this right @olendorf ?

awead:
@olendorf word on the street is that we can, and should: https://github.com/duke-libraries/ezid-client/issues/83

carolyncole:
We need a switch to be able to use ezid, or turn it off completely, or use the new datacite.

carolyncole:
@olendorf  Were you able to get the mapping from our resource Types to the new DataCite types?  We know that Image does work and book does not work, but are not certain of the other types.

carolyncole:
Here are a list of the ResourceTypes from DataCite: https://schema.datacite.org/meta/kernel-4.1/doc/DataCite-MetadataKernel_v4.1.pdf

Audiovisual, Collection, DataPaper, Dataset, Event, Image, InteractiveResource, Model, PhysicalObject, Service, Software, Sound, Text, Workflow, Other

Here is the mapping that will need to be updated: https://github.com/psu-stewardship/scholarsphere/blob/develop/app/services/doi_service.rb#L6

cam156:
Looking at travis builds Master was 18 minutes for the features and 43 minutes for the unit tests, After the "upgrade" to ChromeDriver we were at 34 minutes for the features and 47 minutes for the unit tests. Develop is now 18 minutes for the features and 36 minutes for the unit tests.

[...]

PR #1495, opened 2018-11-20
Use DataCite to mint DOIs

awead:
Updates the ezid-client gem to support the EZ API and adds the required credentials for authentication.

We can still continue to use the current EZID API by using the appropriate `EZID_HOST` and corresponding user, password, and handle. When the time comes to switch to using DataCite's EZ API, we specify the DateCite hostname, as well as corresponding user, password, and new handle.

Connected to #1237


app/services/doi_service.rb
```diff
@@ -24,10 +24,10 @@ class DOIService
     'Other' => 'Other'
   }.freeze
 
-  def initialize(handle = ScholarSphere::Application.config.doi_handle,
-                 user = ScholarSphere::Application.config.doi_user,
-                 password = ScholarSphere::Application.config.doi_password)
-    @client = Ezid::Client.new(user: user, password: password)
+  # @note user, password, and host values for the Ezid::Client are obtained from environment variables
+  #   set in application.yml. The hostname can be either an EZID API service or the new DataCite EZ API.
+  def initialize(handle = ScholarSphere::Application.config.doi_handle)
```
carolyncole, line 29: @awead how does this transition to DataCite?  Does the EzidClient just work?

carolyncole (changes requested):
@awead It would be nice to have a test that calls out to datacite to be certain that it will work when we need to transition.

app/services/doi_service.rb
awead: The ezid-client gem supports both. Just changing the hostname is all that's needed.

awead:
@cam156 should we have a test that's mocked via VCR/webmock?

carolyncole:
@awead a vcr test seems like a good idea.  Did we need to do anything related to the fact that DataCite will not accept Book as a type?

awead:
@cam156 RE: VCR, I agree it would be helpful, but I'm hesitant to do it now given the changes to their test instance. Also, the VCR test wouldn't catch these changes in the future. It would however, ensure that nothing was breaking on our side, so it's still worthwhile. I'd just rather wait until we know for sure what kind of request we'll need to send to them.

awead:
@cam156 BTW, I have tested this branch locally using the current EZID host and credentials that are in production, and DOIs are getting minted. This same branch worked _last week_ before the apparent changes to DataCite's test instance. So we can be confident that this change does not break current production, but due to DataCite's changes, we'd likely have to make additional changes to mint DOI successfully.

That being said, this PR does provide "support" for both EZID and DataCite. It's just that the latter does not accept all our metadata in its present state.

carolyncole (approved):

PR #1540, opened 2019-02-20
Fixes the broken queries for depositor works and collections from the dashboard.

mtribone:
Grabs the _vitals partial from Sufia and overrides depositor field for depositor_ssim, so that the queries from the dashboard return works and collections made by the depositor.

screen shot 2019-02-19 at 8 00 40 pm

screen shot 2019-02-19 at 8 01 11 pm


awead:
@mtribone Looks great. Could you include a one-line comment at the top that mentions we're overriding this from Sufia?

PR #1542, opened 2019-02-25
Added a class to process the Creator spreadsheet

carolyncole:
The class will read in a file and create a list of items to be processed, either split, combined, or modified.  It will also process each item from the list.


spec/lib/clean_agent/clean_agent_spec.rb
```diff
@@ -0,0 +1,174 @@
...
+    let(:agent1) { create :agent, sur_name: 'Sally B Brown' }
+    let(:agent_alias1) { create :alias, display_name: 'Sally B Brown', agent: agent1 }
+    let!(:work1) { create :work, creators: [agent_alias1] }
+    let(:agent2) { create :agent, sur_name: 'Brown', given_name: 'Sally B' }
+    let(:agent_alias2) { create :alias, display_name: 'Sally Brown', agent: agent2 }
+    let!(:work2) { create :work, creators: [agent_alias2] }
+
+    it 'combines the agents' do
+      expect { described_class.combine_agents(agent: agent1, other_agents: [agent2]) }.to change(Agent, :count).by(-1).and change(Alias, :count).by(0)
+      expect { Agent.find(agent2.id) }.to raise_error(Ldp::Gone)
```
awead, line 42: @cam156 Is `agent1` updated to reflect `{sur_name: 'Brown', given_name: 'Sally B' }`, or is it just left as `{ sur_name: 'Sally B Brown' }`? I was expecting the agent name to have their first and last names parsed out correctly as well, but perhaps that not the case here.

awead (approved):

spec/lib/clean_agent/clean_agent_spec.rb
carolyncole: @awead we are just combining agents, so no name changes are done here. Agent 1 gets all aliases that were on Agent 2 and agent 2 gets deleted.

PR #1545, opened 2019-02-27
Making the admin group configurable

carolyncole:
We are planning to change the group name so I'm making it configurable so that is easier


awead (approved):

awead:
@cam156 Looks good. Make sure to update the DLT docs page with the config details for each environment: http://sites.psu.edu/dltdocs/?p=3521

Issue #1324, opened 2018-07-19
Upgrade to ruby 2.4 or later


carolyncole:
This upgrade is needed for compatibility with calmav and the Fedora operating system

It is also needed for the end of life of 2.3.3 in March 2019

2.5 would be nice

carolyncole:
Tried to upgrade to 2.5.1, but sufia is not compatible.

carolyncole:
Ruby 2.4.4 also gets the same error:

/Users/cam156/.rbenv/versions/2.4.4/lib/ruby/gems/2.4.0/gems/therubyracer-0.12.2/lib/v8/conversion.rb:21: warning: constant ::Fixnum is deprecated
bundler: failed to load command: rspec (/Users/cam156/.rbenv/versions/2.4.4/bin/rspec)
Bundler::GemRequireError: There was an error while trying to load the gem 'sufia'.
Gem Load Error is: wrong argument type Class (expected Module)
Backtrace for gem load error is:
/Users/cam156/.rbenv/versions/2.4.4/lib/ruby/gems/2.4.0/gems/therubyracer-0.12.2/lib/v8/conversion.rb:23:in `include'
/Users/cam156/.rbenv/versions/2.4.4/lib/ruby/gems/2.4.0/gems/therubyracer-0.12.2/lib/v8/conversion.rb:23:in `block (2 levels) in <main>'
/Users/cam156/.rbenv/versions/2.4.4/lib/ruby/gems/2.4.0/gems/therubyracer-0.12.2/lib/v8/conversion.rb:22:in `class_eval'
/Users/cam156/.rbenv/versions/2.4.4/lib/ruby/gems/2.4.0/gems/therubyracer-0.12.2/lib/v8/conversion.rb:22:in `block in <main>'
/Users/cam156/.rbenv/versions/2.4.4/lib/ruby/gems/2.4.0/gems/therubyracer-0.12.2/lib/v8/conversion.rb:21:in `each'
/Users/cam156/.rbenv/versions/2.4.4/lib/ruby/gems/2.4.0/gems/therubyracer-0.12.2/lib/v8/conversion.rb:21:in `<main>'
/Users/cam156/.rbenv/versions/2.4.4/lib/ruby/gems/2.4.0/gems/bootsnap-1.2.1/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:21:in `require'
[...]

PR #1579, opened 2019-05-09
Downgrade to Ruby 2.4.6

awead:
Ruby 2.6 was creating problems with cloud upload that wasn't apparent until regression testing. Ruby 2.5 had a different issue with CSV encoding: https://github.com/ruby/csv/issues/23

We're trying out 2.4 to see if it's any better, but it may still have this issue.

Ruby 2.4 EOL date: 2020-03-31

Connected to #1324


awead:
After running the cloud upload regression test, this works. 2.6 was not working with cloud upload, so we'll just have to use 2.4 until we figure out what's going on in 2.6. We have a year...

PR #1592, opened 2019-07-08
updates fits conf. fixes #1587

### LINKED ISSUE #1587: README.md files sometimes identified as video/m2ts
README.md files are sometimes identified as Mime Type: video/m2ts. Behaviour seems to vary, but here is an example:
https://scholarsphere.psu.edu/concern/parent/bnz805z66z/file_sets/mpg15bf61m

whereismyjetpack:


awead (approved):

travis/test.sh
```diff
@@ -18,6 +18,10 @@ fi
 export PATH=$PATH:$(pwd)/dep_cache
 
 echo -e "\n\n\033[1;33mConfiguring Scholarsphere for test\033[0m"
+curl -Lo fits.zip https://projects.iq.harvard.edu/files/fits/files/fits-1.3.0.zip
+# TODO fix this
```
awead: It's unclear what needs to be fixed.

entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
```
awead: is this file related to the fits changes? If so, it could use some comments explaining how. Otherwise, it might need to be in another PR.

.travis.yml
```diff
@@ -43,5 +43,5 @@ jobs:
       script: ./travis/coverage.sh
     - stage: niftany
       script: bundle exec niftany
-after_failure:
-  - cat log/test.log
+# after_failure:
+#   - cat log/test.log
```
awead: Can you uncomment these?

awead (changes requested):
@whereismyjetpack I left a couple of comments. You should also squash all the commits down to one.

travis/test.sh
whereismyjetpack: i'm removing the fits directory, I think we might want to remove the submodule, and then create the directory here instead of needing to remove it

entrypoint.sh
whereismyjetpack: whoops. this shouldn't be here. sorry

.travis.yml
whereismyjetpack: yup!

whereismyjetpack:
@awead I think this is ready to go. branch passed/ presumably the PR will as well, I can squash in the Merge interface when approved.

awead (approved):

PR #1603, opened 2019-07-24
Adds Opt Out to User profile

mtribone:
Adds a checkbox to the user profile and updates the email content to reflect Cynthia's changes. New column is added to the User table for the Boolean. Adds code to the jobs mailer to look for the value of the opt out checkbox and either sends or does not send the email based on user choice.

TODO:
- [x] checkbox set to true, user exists, and sends email
- [x] checkbox set to false and does not send email

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -35,6 +35,11 @@ class UsersController < ApplicationController
     end
   end
 
+  def user_params
+    params.require(:user).permit(:avatar, :facebook_handle, :twitter_handle,
+                                 :googleplus_handle, :linkedin_handle, :remove_avatar, :orcid, :opt_out_stats_email)
+  end
+
   protected
 
     def base_query
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -35,6 +35,7 @@ class UsersController < ApplicationController
     end
   end
 
+  # Overriding the user_params from Sufia, so that we can add the opt_out_stats_email attribute in the user profile
   def user_params
     params.require(:user).permit(:avatar, :facebook_handle, :twitter_handle,
                                  :googleplus_handle, :linkedin_handle, :remove_avatar, :orcid, :opt_out_stats_email)
--- a/app/jobs/user_stats_notification_job.rb
+++ b/app/jobs/user_stats_notification_job.rb
@@ -3,7 +3,7 @@
 class UserStatsNotificationJob < ApplicationJob
   def perform(id:, start_date:, end_date:)
     user = User.find(id)
-    return unless PsuDir::LdapUser.check_ldap_exist!(user.login)
+    return unless PsuDir::LdapUser.check_ldap_exist!(user.login) && User.where(opt_out_stats_email: true)
 
     UserMailer.user_stats_email(
       user: user,
--- a/app/jobs/user_stats_notification_job.rb
+++ b/app/jobs/user_stats_notification_job.rb
@@ -3,7 +3,7 @@
 class UserStatsNotificationJob < ApplicationJob
   def perform(id:, start_date:, end_date:)
     user = User.find(id)
-    return unless PsuDir::LdapUser.check_ldap_exist!(user.login) && User.where(opt_out_stats_email: true)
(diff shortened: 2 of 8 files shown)


app/jobs/user_stats_notification_job.rb
```diff
@@ -3,7 +3,7 @@
 class UserStatsNotificationJob < ApplicationJob
   def perform(id:, start_date:, end_date:)
     user = User.find(id)
-    return unless PsuDir::LdapUser.check_ldap_exist!(user.login)
+    return unless PsuDir::LdapUser.check_ldap_exist!(user.login) && User.where(opt_out_stats_email: true)
```
awead: @mtribone since you have  the `user` object from line 5, you should be able to do `user.opt_out_stats_email` which would return true or false. Putting that together, you can have:
``` ruby
return unless PsuDir::LdapUser.check_ldap_exist!(user.login) && user.opt_out_stats_email
```

app/controllers/users_controller.rb
```diff
@@ -35,6 +35,11 @@ def copy_orcid_id
     end
   end
```
awead, line 37: @mtribone might ought to put a comment here that we're overriding this because of the opt_out_stats_email attribute.

spec/jobs/user_stats_notification_job_spec.rb
```diff
@@ -7,11 +7,12 @@
   let(:start_date) { Date.today - 10 }
   let(:end_date) { Date.today }
 
-  context 'when the user exists in LDAP' do
+  context 'when the user exists in LDAP and wants to receive stats email' do
     let(:mock_mailer) { instance_double(ActionMailer::MessageDelivery) }
 
     before do
       allow(PsuDir::LdapUser).to receive(:check_ldap_exist!).with(user.login).and_return(true)
+      allow(user).to receive(:opt_out_stats_email).and_return(false)
```
awead: @mtribone this is completely fine to do, but I will also point out you could create a new user with this value set like:
`create(:user, opt_out_stats_email: false` and likewise do it with a `true` setting in the other test.

awead:
@mtribone this is looking good to go to me... I put in a few comments to suggest some possible alternatives.

app/jobs/user_stats_notification_job.rb
mtribone: Hmmm...I get failures when I change this line.

awead (approved):

app/views/user_mailer/user_stats_email.html.erb
```diff
@@ -1,15 +1,32 @@
...
-<h1><%= t('statistic.user_stats.heading', date: Date.today.last_month.strftime('%B')) %></h1>
-
-<p>
-  Dear Author, <br />
-  You had <%= @presenter.file_downloads %> new downloads last month across your <%= @presenter.total_files %> files in <%= t('sufia.product_name') %>.
-  You can <a href="https://scholarsphere.psu.edu/dashboard/">view your overall statistics in <%= t('sufia.product_name') %></a> by accessing your dashboard.
-</p>
-
+  <h1><%= t('statistic.user_stats.heading', date: Date.today.last_month.strftime('%B')) %></h1>
+   <p>Dear <%= @given_name %>, <br />
```
whereismyjetpack, line 7: use alias, instead of given name?

app/views/user_mailer/user_stats_email.html.erb
```diff
@@ -1,15 +1,32 @@
...
+    <li>Software and analysis code</li>
+    <li>Book chapters: Check your publishing agreement to see if you can deposit a contributed chapter.</li>
+    <li>Books: Consider
+      <a href="https://www.authorsalliance.org/resources/rights-reversion-portal/">regaining copyright</a>
+      from your publisher so you can deposit an out-of-print book. You can also deposit any book that is openly licensed,
+      such as an open access monograph bearing one of the Creative Commons licenses.
+    </li>
+  </ul>
+  <p>If you have questions please <a href="https://scholarsphere.psu.edu/contact">contact us!</a></p>
 </body>
```
whereismyjetpack, line 31: add opt out instructions / link?

PR #1626, opened 2019-10-04
Hide batch upload links

### LINKED ISSUE #1619: Hide Batch Upload
Batch Upload is prone to misuse. Mainly, it's used to create many single-file works (subsequently gathered under a collection) in situations that call for one work with many files. Also, works created through Batch Upload are titled after the filename, which is a bad practice (i.e. we don't want these kinds of titles indexed by Google Scholar).

While Batch Upload is useful for some situations—for example, when someone is trying to upload PDFs of many separate articles—I don't think it should be emphasized as much as it is in the UI. A link to Batch Upload is one of the first things you see in the "Add New Work" page. 

- remove Batch Upload link from top of New Work Page
- Possibly, Indicate it somewhere else (where?).

srerickson:
Hides the urls for uploading via batch. It's still possible to do it, you just have to know the url now.

Fixes #1619


app/views/_toolbar.html.erb
```diff
@@ -33,7 +33,7 @@
                             ) %>
                   </li>
               <% end %>
-              <li><%= link_to t('sufia.toolbar.works.batch'), sufia.new_batch_upload_path(payload_concern: 'GenericWork') %></li>
+              <li class="hidden"><%= link_to t('sufia.toolbar.works.batch'), sufia.new_batch_upload_path(payload_concern: 'GenericWork') %></li>
```
awead: @srerickson what if we just removed the text completely? I think `class="hidden"` might still be read by screen readers. I'm not sure though.

awead (changes requested):

awead (approved):

PR #1639, opened 2020-06-26
Don't hard-code future dates

awead:
Well, that didn't work out as well as you thought it would, did it now:

https://github.com/psu-stewardship/scholarsphere/commit/6b0a84465731e7cf03aecb39a972a1e514c68e55#diff-ae57f13690d6dca07faad047551cc550R33


jeremyf:
Oh past self…always such a turkey

